PR #5, opened 2021-04-20
fix the key name of the array of abilities-v1 json objects

vu3rdd:
Transit message should look like this..

```
"transit": {
        "abilities-v1": [
            {
                "type": "<string, one of {direct-tcp-v1, relay-v1, tor-tcp-v1}>"
            }
        ],
        ....
```
The key in the arrays is `type` not `kind`.


piegamesde:
Is this a change in the protocol or is this a fixup of its documentation?

vu3rdd:
> Is this a change in the protocol or is this a fixup of its documentation?

Documentation.

PR #7, opened 2021-05-05
Reword and clarify client to client protocol

piegamesde:
Mentions of details that are specific to the python implementation got removed and replaced by a description of the intended semantics. Also added more examples for a lot of the transferred messages. This is a significant shift towards a more implementation-specific target audience.


piegamesde:
This should be mostly uncontroversial, as I didn't do any major modifications.

One thing I need feedback with though is the struct of the `version` phase:

- Which keys are actually described in the dict apart from `app_versions`?
- Which keys go into the content of `app_versions`?
- What is the destinction between the outer level and the content of `app_versions`? What should go where, and who decides?

client-protocol.md
```diff
@@ -3,64 +3,93 @@
...
+In the `version` phase, the clients exchange information about themselves in
+order to do feature negotiation. 
+
+This allows the two Wormhole instances
 to signal their ability to do other things (like "dilate" the wormhole). The
 version data will also include an `app_versions` key which contains a
 dictionary of metadata provided by the application, allowing apps to perform
-similar negotiation.
+similar negotiation. At the moment, no keys other than `app_versions` are
+specified.
```
meejah, line 70: If dilation is enabled in a client, they send a ``can-dilate`` key specifying the dilation versions they support. This is just `"1"` right now AFAICT so this will look like ``{"can-dilate": ["1"]}`` I believe.

piegamesde: \*opens can of worms\*

I'd like to add a `abilities` list in order to promote features. This would allow easy handling by the implementations: simply take the union of ours and their set. This, however, somewhat of clashes with the concept of adding a `can-*ability*` field to each ability.

I think the dilation is not set in stone nor actively used in production, so my preferred approach would be to partially or fully enhance/replace `can-dilate` with the abilities list.

There a few other ways of going about this, but the main alternative is to add `can-*ability*` for other features as well.

piegamesde: I just added a commit of I think how it could work

meejah: Maybe you're right that it's best to just sort of ignore this for now .. Dilation isn't "official" (AFAIK) and so we could add appropriate words here along with some (future) concrete Dilation proposal. The `can-dilate` thing is just how it happens to work now (and the only implementation is the Python one still, AFAIK).

meejah (commented):
Overall, this looks like some solid improvements. I expressed some edits and keeping the "ordering" paragraph as a PR-to-this-PR.

The difference between "versions" and "app_versions" is this: "versions" is only seen by the library. "app_versions" is specified by the application (and is delivered to the other side intact). For example, the Python API lets you specify `versions={}` in the `wormhole.create()` call .. this dict is delivered via `app_versions` and will appear to the other client.

I also left some "bare" comments about maybe ignoring the abilities stuff until a more-cohesive dilation (or seeds) proposal / PR ...? (I could go either way on this .. one advantage of leaving `can-dilate` for now or even just removing it is that there's one implementation that does that, versus zero that do "abilities")

dilation-protocol.md
```diff
@@ -11,15 +11,12 @@ connection has been lost, and then coordinate the construction of a
...
-The Wormhole protocol includes a `versions` message sent immediately after
-the shared PAKE key is established. This also serves as a key-confirmation
-message, allowing each side to confirm that the other side knows the right
-key. The body of the `versions` message is a JSON-formatted string with keys
-that are available for learning the abilities of the peer. Dilation is
-signaled by a key named `can-dilate`, whose value is a list of strings. Any
-version present in both side's lists is eligible for use.
+## Capability discovery
+
+In the `versions` phase of the protocol, the flag `dilation-v1` is added to
```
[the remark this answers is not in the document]
piegamesde: I'm fine with either. What needs to come across in the end is "if you want to do dilation, then add this value to this list".

client-protocol.md
```diff
@@ -60,17 +60,24 @@ encoding is the content of the `body` field.
...
-This allows the two Wormhole instances
-to signal their ability to do other things (like "dilate" the wormhole). The
-version data will also include an `app_versions` key which contains a
+The optional `abilities` key allows the two Wormhole instances
+to signal their ability to do other things (like "dilate" the wormhole).
+It defaults to the empty list. Both sides intersect their abilities with their
+peer's ones, in order to determine wich protocol extensions will be used. An
+ability might define more keys in the dict to exchange more detailed information
+about that feature apart from "I support it". Currently specified abilities are:
+`dilation-v1`, `seeds-v1`.
```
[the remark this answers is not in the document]
piegamesde: How about formulating this as a "list of reserved keywords"?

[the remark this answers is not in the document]
meejah: Sure, yeah I think that's a good approach: we "know" this is coming along, but don't have to think too hard about it (yet) ;)

piegamesde:
I'm ready for (squash and) merge if you are.

meejah (approved):
LGMT, thanks!

PR #10, opened 2021-05-24
document websocket support in the relay server

vu3rdd:


meejah (approved):
LGTM

transit.md
```diff
@@ -48,7 +48,8 @@ This may be relaxed in the future, much as Wormhole was.
 
 Each Transit object has a set of "abilities". These are outbound connection
 mechanisms that the client is capable of using. The basic CLI tool (running
-on a normal computer) has two abilities: `direct-tcp-v1` and `relay-v1`.
+on a normal computer) has these abilities: `direct-tcp-v1`, `relay-v1`,
+`tor-tcp-v1`, `direct-ws-v1` and `direct-wss-v1`.
```
bryanchriswhite, line 52: I feel like these should be called `relay-ws-v1` and `relay-wss-v1` but I also feel like we had this conversation before and I don't recall how it ended. Could you remind me what's up with that? 

If we did go with the above, would it then make sense to rename `relay-v1` to `relay-tcp-v1` (in a backwards compatible way w/ deprecation warning)?

Perhaps we can add descriptions of these in the list that follows? Maybe something like "indicates that it can connect via the transit relay using <proto> transport".

transit.md
```diff
@@ -163,15 +165,19 @@ attempting to use the relay. If it has no viable direct hints, it will start
...
-provided by the Transit instance include the locally-configured relay, along
-with the dynamically-determined direct hints. Both should be delivered to the
+The Transit client attempts connections to multiple relays, and uses
+the first one that passes negotiation. Each side combines a
+locally-configured hostname/port (usually "baked in" to the
+application, and possibly hosted by the application author) with
+additional hostname/port pairs that come from the peer. This way
+either side can suggest the relays to use. The `wormhole` application
+accepts configuration switch that takes as input, the relay server's
+address, for eg: `tcp:myrelay.example.org:12345` or
```
bryanchriswhite, line 175: My understanding is that `e.g.` means ["for example"](https://en.wikipedia.org/wiki/List_of_Latin_phrases_(E)#e.g.), making the "for" redundant.
```suggestion
address, e.g.: `tcp:myrelay.example.org:12345` or
```

transit.md
piegamesde, line 52: I know it's kind of tangential to your change, but why would one advertise any of the other three values? What meaning do they have in the protocol/handshake? I think the only relevant question is "can we connect directly or do we need a relay"?

vu3rdd: What you are saying makes sense, now that relay specification also includes the protocol. It wasn't there in the past before the WebSocket changes.

vu3rdd: Perhaps those changes should be separate from this change as a "-v2" abilities? That would also help maintain backward compatibility with older relay servers and clients.

piegamesde:
I'd like to discuss the semantics of having multiple protocols towards one relay server. What is the intended feature negotiation once both sides have exchanged their abilities? Usually, I'd say both sides have to agree on each (i.e. take the intersection of both sets), but with web sockets this needs not to be so: two clients could connect through a relay, one with web sockets and one over TCP. (Corollary: do both sides actually need to use TOR in order to use TOR or is only one side sufficient if additionally a relay is used? I don't know how TOR works internally at all.)

bryanchriswhite:
@piegamesde I can't speak to tor because I too am not well-read on tor. Additionally, [wormhole-william](https://github.com/psanford/wormhole-william), which we're using for the current project, hasn't implemented tor support yet.

I'm not aware of any reason that the relay wouldn't be able to speak a different protocol on each leg. @meejah implemented the transit relay side of the change so perhaps he has more concrete details here.

The original motivation for us to add a WebSocket as a transport option was to support a web environment. In this context, use of Websocket transport implies use of the relay (as browsers can't listen for incoming connections*).  In a scenario between clients which both support TCP and WebSocket transport, the relay is not necessarily a given; thus, I think the following would apply from the [transit spec](https://github.com/magic-wormhole/magic-wormhole-protocols/blob/main/transit.md#handshake-direct-tcp-v1-tor-tcp-v1):
> To tolerate the inevitable race conditions created by multiple contending sockets, only the Sender gets to decide which one wins: the first one to make it past negotiation. Hopefully this is correlated with the fastest connection pathway. The protocol ignores any socket that is not somewhat affiliated with the matching Transit instance.

bryanchriswhite:
@piegamesde in a scenario where both peer clients support multiple transports and the relay is used, I think it makes sense for the client to only connect to the relay with one transport at a time.

Good observation!

piegamesde:
I don't really know how to move this forward. I also kind of depends on how willing we are to make big breaking changes to the relay protocol. The simplest solution I can see is to add the client side in the relay handshake:

`please relay $token for $side\n`

That way the relay will know if two connection attempts stem from the same entity and can safely ignore the all except the first one. Note that this is not 100% backwards compatible on the Relay side, but at the moment there is only one and it is in our control so I don't feel that bad about it.

While it solves the problem with the race hazard, it still does not resolve the other problems I see with the current abilities merging: In theory, an application supporting only websockets can communicate with an application supporting only TCP just fine, provided that the relay supports both. But there isn't really a way to tell whether a relay does, so it may just horribly fail.

meejah:
The relay server already uses `please relay X for side Y` (although it still works if you use `please relay X` .. because that's the original protocol).

meejah:
For Tor (which is not an acronym so they prefer "Tor" over "TOR"): with client-to-server situations, a Tor connection can be thought of as just a normal TCP connection except that the server doesn't know where the original client is on the internet. So, either client can use Tor to connect to the relay server (for example) independently of each other.

An even better way to use Tor would be for one side to set up an Onion service and advertise it to the other. An Onion service is a Tor way to be a server. Clients of such a server have to use Tor. In this setup, the relay server would not be needed but there also won't be any NAT issues because all connections to the Tor network are "outgoing" (including for Onion services). (We don't yet support this mode though).

meejah:
The current relay implementation can handle either TCP or WebSockets on either side (so can relay WS-to-WS or TCP-to-WS on either leg).

piegamesde:
Thank you for the input. In that case I suggest the following changes to the protocol:

- Document the `please relay X for side Y` and what this means for relay servers
- Document that self-loops may happen when not specifying the side
- Keep the abilities set to `direct-tcp-v1` and `relay-v1` (do we need to bump the version here?). I thought about having more flags and only sending relay URLs to the peer that they'd support. But then I figured that input sanitation will needed to be done anyways, so I don't see the point in doing this. Maybe reserve `tor-hidden-v1` for "I am able to spin up a hidden service" or something.

meejah:
I think all that sounds good @piegamesde .. it might be that we can side-step the "loops" problem: any client new enough to support WebSockets should also support the `for side Y` part of the handshake. And I _think_ the only case where a loop might happen is when a client is using TCP plus some other kind (right?).

That said, we should probably still document the issue (and that new clients should always use the `for side` variant).

piegamesde:
> And I think the only case where a loop might happen is when a client is using TCP plus some other kind (right?).

That's the obvious case, but no. It may end up happening whenever a client has two entries with different content to the same server. Yes, I don't expect this to ever happen, but I still think it should be documented that it *might* happen as it clearly is not impossible.

> we could enforce this by insisting that WebSocket clients can only use the please relay X for side Y handshake...

It'd only partially help, but it's a good idea nevertheless.

bryanchriswhite:
> The relay server already uses `please relay X for side Y` (although it still works if you use `please relay X` .. because that's the original protocol).

I went looking for this in the spec because I thought i recalled something about "for side ..." being a thing but didn't see it. Was this a change that just didn't get documented?

meejah:
@bryanchriswhite maybe? That handshake was already in there when I implemented WebSocket support ...

piegamesde:
It was already there when @vu3rdd implemented transit in the Rust port a few years ago. But I've bisected the original commit with that change, it's from the end of 2016 and it contains no further motivation: https://github.com/magic-wormhole/magic-wormhole/commit/e1546bf03fc404ecdb6543f57835ec77a609c247

piegamesde:
Coming back to this again, this time with from the perspective of adding UDT transport (personal experiments) and advertising relay servers in the welcome message. Some random thoughts:

- Clarify the relay handshake as discussed above
- The abilities exchange should be a separate step from the hints exchange, because the hints require some setup based on the peer's abilities. The spec already kind of mentions this, but it should be made more clear. This won't help us right now since there is no way to change the transfer protocol to respect this, but an eventual transfer v2 will have to correct that mistake.
- Relay servers (and thus, relay hints) are advertised with all supported protocols, independently of the other abilities (if the relay supports tor, include its onion address for example). The hints are bound to the server so that only one connection needs to be made per server. Hints can be as simple as an encoded list of URIs, where the schema determines the protocol.
- The tricky bit is to update the current transfer protocol to work with this. I suggest adding a new `relay-v2` ability and deprecate the old.
  - Older clients ignore the `relay-v2` hint
  - Newer clients must send a `relay-v1` hint alongside the v2 that includes the encodable subset of the information (notably, the server's TCP address if present).
  - Newer clients ignore the v1 hint if a v2 hint is present
  - If abilities are sent separately from hints (sadly not our case right now), and both clients signal a `relay-v2` ability, then the v1 compat may be skipped.
- General protocol clarification: clients MUST support relay servers, and they SHOULD send that ability. The only acceptable exception is when a direct connection attempt is being enforced. On the other hand, clients don't need to support any direct connections, relying solely on the relay. Clients that wish to hide their IP address SHOULD NOT use `direct-*` hints.

If you want and find this proposal acceptable, I can draft a PR containing the spec clarifications.

piegamesde:
At the moment, only the first mentioned use case is actually implemented by clients. Use case two is covered by the `tor-tcp-v1` hint, and thus is out of scope of this discussion. Use case three is redundant to the other two in my opinion, but my guess at specifying this would be to add a `tor-relay-v1` ability & hints. (Actually, this is almost exactly the same thing as the second, except that the clients first do a relay handshake upon connecting.)

bryanchriswhite (approved):

PR #14, opened 2021-07-10
Add changelog for 1.0.1

piegamesde:
Not sure if we want to also document changes on the repository itself that are not part of the spec (like build infrastructure and the changed communication channel), but I'm slightly against it.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,11 @@
 
 ## Unreleased chnages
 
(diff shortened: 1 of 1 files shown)


meejah (approved):
LGTM

I'm neutral on whether "meta" changes like IRC / repo-stuff need to be mentioned. This is "a spec" so that seems like noise in the changes to me too?

PR #16, opened 2021-09-22
Transit: protocol improvements

piegamesde:
A prototype implementation can be found [here](https://github.com/magic-wormhole/magic-wormhole.rs/pull/130).

Also note that stun.magic-wormhole.io doesn't exist yet ^^

### DIFF
--- a/transit.md
+++ b/transit.md
@@ -142,9 +142,13 @@ magic-wormhole instances. It uses a TCP-based protocol with a handshake to
 determine which connection wants to be connected to which.
 
 When connecting to a relay, the Transit client first writes RELAY-HANDSHAKE
-to the socket, which is `please relay %s\n`, where `%s` is the hex-encoded
-32-byte HKDF derivative of the transit key, using `transit_relay_token` as
-the context. The client then waits for `ok\n`.
+to the socket, which is `please relay $channel for $side\n`, where `$channel`
+is the hex-encoded 32-byte HKDF derivative of the transit key,
+using `transit_relay_token` as the context, and `$side` is a random per session
+identifier. The `side` is used to deduplicate a client opening multiple
+connections to the same relay server: without, it may result in a loopback
+to itself and a dead-lock of the protocol.
+The client then waits for `ok\n`.
 
 The relay waits for a second connection that uses the same token. When this
 happens, the relay sends `ok\n` to both, then wires the connections together,
--- a/transit.md
+++ b/transit.md
@@ -30,7 +30,7 @@ both connect to a relay server.
 
 ## Roles
 
-The Transit protocol has pre-defined "Leder" and "Follower" roles (unlike
+The Transit protocol has pre-defined "Leader" and "Follower" roles (unlike
 Wormhole, which is symmetric/nobody-goes-first), which are called "Sender" and "Receiver"
 for historical reasons. Each connection must have exactly one Sender and exactly one Receiver.
 If the application using transfer does not provide this distinction, some deterministic
@@ -176,7 +176,7 @@ peer.
 
 ## Encryption
 
-If desired\*, transit provides an enrypted **record-pipe**, which means the two
+If desired\*, transit provides an encrypted **record-pipe**, which means the two
 sides can and receive whole records, rather than unframed bytes. This is a side-effect of the
 encryption (which uses the NaCl "secretbox" function). The encryption adds 44
 bytes of overhead to each record (4-byte length, 24-byte nonce, 32-byte MAC),
@@ -217,3 +217,111 @@ both sides have their own nonce.
 For each received record, the nonce must be checked to be equal to the expected value,
 which has to be tracked. Thus, there are 4 nonces in total, two per direction which are
 kept in sync.
+
+## On finding a direct connection
+
+Because it is easy to get things wrong. This section only applies to the
+`direct-tcp-v1` hint. The following section is closely tied to the operating
+system's sockets API, and the the verbs *bind*, *listen*, *accept* and *connect*
+will be used accordingly. Note that this is more a guide than a specification,
+since it handles implementation details that should all mostly be compatible
+with each other (and when not, failure is not critical).
+
+### The easy way
+
+Each side binds a new TCP socket, enumerates all local non-loopback IP addresses
+and exchanges port+IPs with the peer. Then, both sides listen on their socket
+while simultaneously trying to connect to the peer. As described above, once a
+handshake is successful, all other attempts are aborted and the resources closed.
+
+### Punching holes through firewalls
+
+To go through firewalls, a different approach must be used instead: first, both
+sides again need to bind a socket and exchange hints. This time however, the
+`SO_REUSEADDR` option must be set. Then, both sides try to connect to each of
+the peer's IPs. For every connection attempt, they bind a new socket on the same
+port again. Also you read that right, there are no `listen` or `accept` calls
+involved.
+
+This works only with firewalls that silently filter packets and which don't send
+back any RST packets. Unfortunately, this is the default behavior of the kernel
+when no firewall is active (or the firewall allowed) the packets. This means
+that the firewall hole punching method does *not* work when *no* firewall is
+active (Fun fact: if one side has a firewall and the other side doesn't then
+connection attempts will work or fail depending on which side sent their packets
+first—I'm really glad for you that you didn't have to debug this …). Thus,
+**this method can only be an addition to the "easy" way** specified above. Sadly,
+this means that the number of hints – and thus, attempted connections – is
+doubled. If you manage to listen and accept on the same port so that only one
+set of hints needs to be sent out, please contact `piegames` with your findings.
+Also read [this paper](https://bford.info/pub/net/p2pnat/) (Python code
+[here](https://github.com/dwoz/python-nat-hole-punching/)) for more details on
+the subject.
+
+### Traversing NAT
+
+To additionally traverse through a potential NAT, clients need to query their
+external IP address from a STUN server\*. Because of how NATs work, the socket
+must again be bound with `SO_REUSEADDR` and its port must be used for the hint.
+Clients may `shutdown` the connection to the STUN server, but the socket must
+be kept open. No enhanced NAT detection or other advanced STUN features need to
+be used: we only care about the external IP address and if it fails we can always
+simply fall back to the relay. Generally, a failure to do STUN should be silent
+and the query should time out rather quickly.
+
+Recall that NATs are mainly a hack around an outgrown address space and that IPv6
+users usually have a globally routed public address, for which simply firewall
+hole punching is sufficient. For this reason, you only need to do an IPv4 query
+and you always want to prefer IPv6 connections.
+
+\* Since you should always host your own for production use (and also most of
+the public ones don't support TCP), we provide <tcp://stun.magic-wormhole.io:3478>
+as part of the Magic Wormhole infrastructure (please only use for Magic Wormhole
+clients).
+
+### Choosing the best connection
+
+When multiple connection attempts go through, it is up to the leader side to
+select the one to use and to cancel the other ones. A naive implementation might
+simply use the first one that got established, in the assumption that it will be
+the best (it probably had the best round trip time, modulo some jitter). But next
+to that, there are other criteria that might be taken in on the rating:
+
+1. Prefer direct connections over relay ones
+2. Prefer local connections over those that route through the internet
+3. Prefer IPv6 over IPv4
+4. (the optional `priority` field on the hints should be ignored for now, as it
+  is under-specified to the point of being useless.)
+5. (Similar to point 4., an interface letting the user specify this might be
+  provided.)
+
+The first two points are easy to implement: after a connection established, wait
+some time more to see if a "better" connection shows up. The waiting time should
+be roughly derived from the time that was needed for the first one, but with some
+upper bound to not annoy users.
+
+The third point is a bit fuzzier: the line quickly becomes blurry when VPNs and
+corporate network setups are involved. For IPv4, simply check whether the connection
+was made over the external IP address or if it uses a prefix commonly used for
+local networks. For IPv6, an easy heuristic is to say that a connection where
+both sides' IP addresses have a longer prefix in common is preferable.
+
+### Potential pitfalls and other considerations
+
+- The socket options that must be set are really subtle, for example read
+  [this excellent StackOverflow answer](https://stackoverflow.com/a/14388707/6094756)
+  for more details. In short, you need to check whether the operating system
+  supports reusing ports and which options must be set—on some systems,
+  `SO_REUSEADDR` suffices, others provide `SO_REUSEPORT`. If the system doesn't
+  support it, fall back on the "easy" way and/or relay servers.
+- To prevent port hijacking, make sure that the port may not be reused by other
+  processes. Some systems provide special exclusivity options, while others enable
+  them by default.
+- From the first moment a socket is bound to a port, that socket must be kept
+  open. Simply binding and not using it is okay. If you just bind one to get a
+  free port and then close it immediately, another application may (accidentally)
+  hijack the port until you want to use it again.
+- Generally, all implementations should bind only IPv6 sockets (`AF_INET6`). The
+  kernel will take care of translating the packets to IPv4 if required. That way,
+  no special support is required to handle both stacks. Conversely, only binding
+  to `127.0.0.1` means that the application will only support IPv4.
--- a/transit.md
+++ b/transit.md
@@ -184,7 +184,12 @@ If desired\*, transit provides an encrypted **record-pipe**, which means the two
 sides can and receive whole records, rather than unframed bytes. This is a side-effect of the
 encryption (which uses the NaCl "secretbox" function). The encryption adds 44
 bytes of overhead to each record (4-byte length, 24-byte nonce, 32-byte MAC),
-so you might want to use slightly larger records for efficiency. The maximum
+so you might want to avoid bite-sized records for efficiency reasons. Tranport
+protocols that already include message framing (i.e. WebSockets) omit the four
+byte length prefix of each message. It is up to the relay to add or remove them
+where appropriate.
+
+The maximum theoretical
 record size is 2^32 bytes (4GiB). The whole record must be held in memory at
 the same time, plus its ciphertext, so very large ciphertexts are not
 recommended. Transit implementations must implement a mechanism to set an upper
--- a/transit.md
+++ b/transit.md
@@ -187,7 +187,9 @@ bytes of overhead to each record (4-byte length, 24-byte nonce, 32-byte MAC),
 so you might want to use slightly larger records for efficiency. The maximum
 record size is 2^32 bytes (4GiB). The whole record must be held in memory at
 the same time, plus its ciphertext, so very large ciphertexts are not
-recommended.
+recommended. Transit implementations must implement a mechanism to set an upper
+bound to the message size, which protocols using transit may use. That value should
+default to 64MiB, which is a common default for WebSockets or HTTP communication.
 
 Transit provides **confidentiality**, **integrity**, and **ordering** of
 records. Passive attackers can only do the following:
--- a/transit.md
+++ b/transit.md
@@ -47,32 +47,28 @@ This may be relaxed in the future, much as Wormhole was.
 ## Abilities
 
 Each Transit object has a set of "abilities". These are outbound connection
-mechanisms that the client is capable of using. The basic CLI tool (running
-on a normal computer) has two abilities: `direct-tcp-v1` and `relay-v1`.
+mechanisms that the client is capable of using. The following abilities are
+specified in this document:
 
 * `direct-tcp-v1` indicates that it can make direct outbound TCP connections to a
   requested host and port number.
 * `relay-v1` indicates it can connect to the Transit Relay and speak the
   matching protocol.
-* `tor-tcp-v1`
-
-Future implementations may have additional abilities, such as connecting
-directly to Tor onion services, I2P services, WebSockets, WebRTC, or other
-connection technologies. Implementations on some platforms (such as web
-browsers) may lack `direct-tcp-v1` or `relay-v1`.
+* `relay-v2` is the same as `relay-v1`, but it adds support for an improved hint
+  encoding that allows using WebSockets (and possibly other protocols) for
+  transport towards the relay server.
+* `tor-tcp-v1` allows both sides finding eath other over Tor
 
 Together with each ability, the Transit object can create a
 list of "hints", which tell the respective handshake how to find the other side.
-Hints usually contain a `hostname`, a `port` and an optional `priority`.
-Each hint will fall under one of the abilities that the peer indicated it
-could use. Hints have types like `direct-tcp-v1`, `tor-tcp-v1`, and
-`relay-v1`.
+Each ability declares its own set of hints; hints have a `type` that is equal
+to the name of the ability they hint for.
 
 For example, if our peer can use `direct-tcp-v1`, then our Transit object
 will deduce our local IP addresses (unless forbidden, i.e. we're using Tor),
 listen on a TCP port, then send a list of `direct-tcp-v1` hints pointing at
-all of them. If our peer can use `relay-v1`, then we'll connect to our relay
-server and give the peer a hint to the same.
+all of them. If our peer can use `relay-v1`/`relay-v2`, then we'll connect to
+our relay server and give the peer a hint to the same.
 
 `tor-tcp-v1` hints indicate an Onion service, which cannot be reached without
 Tor. `direct-tcp-v1` hints can be reached with direct TCP connections (unless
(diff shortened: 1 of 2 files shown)


transit.md
```diff
@@ -167,16 +167,71 @@ The Transit client can attempt connections to multiple relays, and uses the
...
+```
+
+Each entry in `hints` is equivalent to a `direct-tcp-v1` hint. Due to historic
+weirdness of the encoding, no more than one single relay server should be hinted at.
+
+### `relay-v2`
+
+```json
+{
+  "type": "relay-v2",
```
meejah, line 219: For UI/UX things it might be useful to have a `"name"` as well? (e.g. if a user might ever be expected to choose one hint/group-of-uRLs over another).

transit.md
```diff
@@ -217,3 +272,108 @@ both sides have their own nonce.
...
+have to debug this …). Thus, **this method can only be an addition to the "easy" way** specified above.
+Sadly, this means that the number of hints – and thus, attempted connections – is doubled. If you manage
+to listen and accept on the same port so that only one set of hints needs to be sent out, please contact
+`piegames` with your findings. Also read [this paper](https://bford.info/pub/net/p2pnat/) (Python code
+[here](https://github.com/dwoz/python-nat-hole-punching/)) for more details on the subject.
+
+### Traversing NAT
+
+To additionally traverse through a potential NAT, clients need to query their external IP address
+from a STUN server\*. Because of how NATs work, the socket must again be bound with `SO_REUSEADDR`
```
meejah, line 313: This might benefit from discussing symmetric vs cone etc etc NATs, or pointing at another resource for this .. there's of course a whole bunch of hole-punching research over the last two decades and we can't possibly be comprehensive here. (Specifically, I don't believe this technique will work for "restricted" NATs that require at least one outgoing packet to the destination .. and some require the right port too). Of course, we don't "need" any particular technique to work since they're "just" hints, but ...

transit.md
```diff
@@ -217,3 +272,108 @@ both sides have their own nonce.
...
+and its port must be used for the hint. Clients may `shutdown` the connection to the STUN server, but
+the socket must be kept open. No enhanced NAT detection or other advanced STUN features need to be
+used: we only care about the external IP address and if it fails we can always simply fall back to
+the relay. Generally, a failure to do STUN should be silent and the query should time out rather
+quickly.
+
+Recall that NATs are a structure of the legacy and that IPv6 users usually have a globally routed
```
meejah, line 320: "of the legacy ..." word missing here? (I guess you mean "Internet" maybe?). I'm not sure that dismissing "the IPv4 Internet" as "legacy" helps the cause here; it'll be around for some time by the looks of things.

transit.md
```diff
@@ -217,3 +272,108 @@ both sides have their own nonce.
...
+
+When multiple connection attempts go through, it is up to the leader side to select the one to use and
+to cancel the other ones. A naive implementation might simply use the first one that got established,
+in the assumption that it will be the best (it probably had the best round trip time, modulo some jitter).
+But next to that, there are other criteria that might be taken in on the rating:
+
+1. Prefer direct connections over relay ones
+2. Prefer IPv6 over IPv4
+3. Prefer local connections over those that route through the internet
+4. (the optional `priority` field on the hints should be ignore for now, as it is under-specified to the
```
meejah, line 338: nit: "ignored".

transit.md
```diff
@@ -217,3 +272,108 @@ both sides have their own nonce.
...
+  point of being useless.)
+5. (Similar to point 4., an interface letting the user specify this might be provided.)
+
+The first two points are easy to implement: after a connection established, wait some time more to see if a
+"better" connection shows up. The waiting time should be roughly be derived from the time that was needed for
```
meejah, line 343: nit: extra "be" ("...should be roughly derived from...")

transit.md
```diff
@@ -217,3 +272,108 @@ both sides have their own nonce.
...
+
+### Choosing the best connection
+
+When multiple connection attempts go through, it is up to the leader side to select the one to use and
+to cancel the other ones. A naive implementation might simply use the first one that got established,
+in the assumption that it will be the best (it probably had the best round trip time, modulo some jitter).
+But next to that, there are other criteria that might be taken in on the rating:
+
+1. Prefer direct connections over relay ones
+2. Prefer IPv6 over IPv4
```
meejah, line 336: For 1 and 3 the justification seems intuitive. It's less clear why v4 vs v6 matters? (At least, I feel this should be "3" since preferring local connections seems more important)

transit.md
```diff
@@ -217,3 +272,108 @@ both sides have their own nonce.
...
+The first two points are easy to implement: after a connection established, wait some time more to see if a
+"better" connection shows up. The waiting time should be roughly be derived from the time that was needed for
+the first one, but with some upper bound to not annoy users.
+
+The third point is a bit fuzzier: the line quickly becomes blurry when VPNs and corporate network setups are
+involved. For IPv4, simply check whether the connection was made over the external IP address or if it uses
+a prefix commonly used for local networks. For IPv6, an easy heuristic is to say that a connection where both
+sides' IP addresses have a longer prefix in common is preferable.
+
+### Potential pitfalls and other
```
meejah, line 351: "and other considerations"?

transit.md
```diff
@@ -217,3 +272,108 @@ both sides have their own nonce.
...
+  requires direct (or relayed) TCP connection
+* WebRTC: usable by web browsers, hard-but-technically-possible to use by
+  normal computers, provides NAT hole-punching for "free"
+* (web browsers cannot make direct TCP connections, so interop between
+  browsers and CLI clients will either require adding WebSocket to CLI, or a
+  relay that is capable of speaking/bridging both)
+* I2P: like Tor, but not capable of proxying to normal TCP hints.
+* ICE-mediated STUN/STUNT: NAT hole-punching, assisted somewhat by a server
+  that can tell you your external IP address and port. Maybe implemented as a
+  uTP stream (which is UDP based, and thus easier to get through NAT).
```
meejah, line 379: proabably worth adding a point about Tor onion services, since that got deleted in a paragraph above. Perhaps:

* Tor Onion services: one side could run an Onion service, allowing a "direct" Tor connection without using any relay

transit.md
```diff
@@ -135,16 +130,21 @@ your wormhole) can cause their peer to make a TCP connection (and send the
...
-determine which connection wants to be connected to which.
+magic-wormhole instances. Clients connect to the relay either via TCP
+or via WebSocket with a handshake to determine which connection wants
+to be connected to which.
 
 When connecting to a relay, the Transit client first writes RELAY-HANDSHAKE
-to the socket, which is `please relay %s\n`, where `%s` is the hex-encoded
-32-byte HKDF derivative of the transit key, using `transit_relay_token` as
-the context. The client then waits for `ok\n`.
+to the socket, which is `please relay $channel for $side\n`, where `$channel`
```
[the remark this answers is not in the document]
piegamesde: I agree that the section is poorly worded. But I'm not sure about the framing bit: the whole thing still works if you ignore the packages and treat it as a continuous byte stream.

transit.md
```diff
@@ -167,16 +167,71 @@ The Transit client can attempt connections to multiple relays, and uses the
...
+      "port": "<tcp port>",
+      "priority": "<number, usually [0..1], optional>"
+    },
+    …
+  ]
+}
+```
+
+Each entry in `hints` is equivalent to a `direct-tcp-v1` hint. Due to historic
+weirdness of the encoding, no more than one single relay server should be hinted at.
```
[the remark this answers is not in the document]
piegamesde: That's the point: I actually don't know. The current specification has one `relay-v1` hint which contains a *list* of direct hints. But since I've never seen more than one hint (in total), I don't know what the semantics are. I'd have to dig up some code. My first guess was that there is only one `relay-v1` hint which contains all the relay servers, which would be different to what is done in `relay-v2`. But again, I don't know.

transit.md
piegamesde: What would the name of such a relay server be? Wouldn't picking a domain name from the URLs suffice? But I can add an optional `name` field if you want.

transit.md
piegamesde: I try to keep this paragraph as "just do this and it'll work most of the time", since this topic is already complex enough. But I'll add a link to the paper that helped me implementing – although I didn't use their algorithm (couldn't get it to work), they go really in depth on all the NAT types and different scenarios and edge cases.

transit.md
piegamesde: Can't wait until my spell checking plugin works again …

transit.md
piegamesde: Oh, did I delete too much somewhere? Everything that the Python implementation does should be covered by the spec.

piegamesde:
Thank you for mentioning the framing, this is a good point: we only require a continuous byte stream (as TCP gives us) because we build our own framing on top. With WebSockets, this seems wasteful. Should I describe a variant that removes the framing for protocols that already bring their own? This would amount to saving the `\n` in the handshake and then the four byte length prefix per message, which is negligible if you only care about saving bytes.

transit.md
```diff
@@ -217,3 +274,125 @@ both sides have their own nonce.
...
+simply fall back to the relay. Generally, a failure to do STUN should be silent
+and the query should time out rather quickly.
+
+Recall that NATs are mainly a hack around an outgrown address space and that IPv6
+users usually have a globally routed public address, for which simply firewall
+hole punching is sufficient. For this reason, you only need to do an IPv4 query
+and you always want to prefer IPv6 connections.
+
+\* Since you should always host your own for production use (and also most of
+the public ones don't support TCP), we provide <tcp://stun.magic-wormhole.io:3478>
```
piegamesde, line 335: By the way we need infrastructure access to set this up (CC @warner). In the meantime, I host a public instance at `stun.piegames.de`.

piegamesde:
I feel a bit weird for bumping the relay version although we technically only change the message formatting and not the way the protocol actually works. I just had an idea for an alternative implementation: the abilities are already serialized as objects, i.e. they may have complex structured values. So instead of having a `relay-v2` ability, we could send `"relay-v1": { "non-tcp-support": true }` or something. The `hints` field would then become a bit of a misnomer and outdated, but I think that's kind of okay. (An open question would be whether the TCP hints are announced via the old `hints` field, which is inconsistent, or with the `urls`, which would make the `hints` redundant and obsolete.)

piegamesde:
:bell: I think this is good to go, so let's start a Final Comment Period! :bell:

I expect all commits except the last one to be rather uncontroversial. If this turns out to be true and the first part does not get any complaints, I'll split this up into two PRs to get that half merged early.

meejah:
Is the intent of "version A" versus "version B" to decide in this PR on one and only use that, or something else?
I also think a complete JSON example with e.g. two different relays and multiple protocols would be good.

piegamesde:
> Is the intent of "version A" versus "version B" to decide in this PR on one and only use that, or something else?

No. The thing is, that due to how our abstractions are layered, that transit cannot specify how the required messages should be encoded. This is part of the application protocol on top. Usually, there is a "canonical" encoding that everybody follows, like for the hints. For the abilities however, it has some (minor) drawbacks that are improved in Version B. That one is not used anywhere at the moment, except by my port forwarding and file transfer v2 experiments.

That being said, I'll remove Version B and mark Version A as "canonical". This will make it easier to understand. It does not change any semantics, since the upper layer protocol still has the last word on how things should be done.

> I also think a complete JSON example with e.g. two different relays and multiple protocols would be good.

Will do.

transit.md
```diff
@@ -176,13 +176,48 @@ provided by the Transit instance include the locally-configured relay, along
...
+
+```json
+[
+  {
+    "type": "direct-tcp-v1"
+  },
+  {
+    "type": "relay-v1",
+    "url_hints": true
+  },
```
meejah, line 209: Now I'm really confused: what is this communicating (i.e. why aren't the actual URLs here)?

transit.md
```diff
@@ -202,6 +237,10 @@ IP address or a domain. Furthermore, IPv6 link-local addresses are not supported
 ```json
 {
   "type": "relay-v1",
+  "name": "<string, optional>",
+  "urls": [
+    "<protocol>://<host>:<port>",
+  ],
```
meejah, line 243: Since we're already communicating structured data here, I really think this would benefit from _not_ trying to follow a single "URL" format (that will need its own parser anyway) and instead look more like:

```
urls: [
    {
        "protocol": "tcp",
        "host": "...",
        "port": 1234
    }
]
```

meejah: ...and maybe `transports` or something is more appropriate than `urls` in that case.

piegamesde: I'd be open to changing this to a more structured format, but how would the WebSockets hint look like? WebSockets paths are already URLs by nature and thus rather complex (a WebSocket can be on an arbitrary path).

The other point is that most implementations will need an URL parser anyways, because that's the most natural way to let users input them. (Also, every WebSockets implementation already has a full-blown HTTP client inside, so the dependencies are already there)

transit.md
piegamesde: Because this is only the *ability*, telling the peer that you are *capable of* sending hints with the new format. Because if you know your peer doesn't, you then can save the effort.

But I'm not entirely sure where the confusion comes from, since this is an existing concept of the protocol that I didn't really change. Maybe it's because the current transfer protocol sends both abilities and hints in a single message (IIRC), which may reduce one round trip but kind of limits the usefulness of a prior capability exchange.

transit.md
piegamesde: Okay, quick alternative proposal for the encoding:

- add a `type` field to `hints` to distinguish different transports, where we assume `TCP` as default if omitted for backwards compatibility.
- a new `type: WebSockets` hint is added: `{ "type": "WebSockets", "url": "ws://example.com" }`. It features only an URL that you can directly feed into the WebSockets library without hazzle. In that same spirit, `ws` and `wss` hints use the same kind of hint.
- The `urls` array is not added anymore
- The `url_hints` in the ability needs to get a different name then. (Maybe `fancy_transport`? :D)

Example:

```json

{
  "type": "relay-v1",
  "name": "Example relay",
  "hints": [
    {
      "type": "TCP",
      "hostname": "example.org",
      "port": "1234"
    },
    {
      "type": "WebSockets",
      "url": "wss://example.org:8000"
    }
  ],
}
```

meejah: I like the look of that a lot better. This would allow expansion into other protocols that may or may not have a well-defined `proto://` style thing (_is_ `tcp:` actually well-defined?) like QUIC, Tor, or others.

For sure websocket transports _do_ involve the whole HTTP stack of course.

I would maybe stick with lower-case only for the type to avoid confusion (which has actually happened in WebSocket things, unfortunately). `"type": "tcp"` or `"type": "websocket"` (and that is defined as "websocket" or "WebSocket" .. i.e. without the "s").

The RFC specs the value as all lower-case when it's a value .. e.g. `Upgrade: websocket` .. and HTTP headers are mixed-case, but IIRC _that_ RFC says they shall be interpreted as case-insensitive.

So for the hints stuff, talking about the version, you meant you prefer the `"url_hints": true` thing to just bumping that to `relay-v2` .. because the actual relay protocol isn't changing, just the hints-exchange method...?

piegamesde: > is tcp: actually well-defined

Kinda? I did some research initially and I found a few applications using it that way, but no official "standard" of any sorts.

Your naming proposals are good.

 > So for the hints stuff, talking about the version, you meant you prefer the "url_hints": true thing to just bumping that to relay-v2 .. because the actual relay protocol isn't changing, just the hints-exchange method...?

Yes. Basically, the change is backwards compatible and clients that support the new features can work together with clients that don't pretty easily. (As far as I checked)

meejah: I'm still kind of split on the version -- having stuff "continue to work" is nice, but the bool flag is .. not nice ;)

piegamesde: What would be your preferred way of handling the versioning? We could add `relay-v2` instead and say that they are identical except for the hints encoding, but I haven't thought any of this through yet.

Also I'd like to know what is your problem with adding additional data to the ability? Is it because it would be difficult to map to language or framework constructs in some way? (I always look at these from the Rust perspective, but Rust has very powerful enums so maybe that's unfair)

piegamesde: I tried out the `relay-v2` approach in the latest commit, let me know what you think.

meejah: I don't have a strong objection to adding things to the abilities .. it's just that a bool like that kind of feels like a "different version"..? Like if you have a function `def decrypt(bool is_aes):` it might be better to have two functions `def decrypt_aes()` and `def decrypt_default()` instead.

I do like that latest commit a lot (a couple typos / nitpicks, see below). Lets think about backwards-compatibility in both cases:

*only `relay-v1`*: implementations select on the abilities `url_hints` to enable support for parsing new-version hints. Peers with `relay-v1` ability but either peer having `url_hints: false` will choose non-URL hints and pass only those.

*`relay-v2` and `relay-v1`*: implementations select parsing based on the abilities string. Peers where either one has no `relay-v2` ability will only pass old-style hints.

Should be fairly similar, implementation-wise. I think I prefer the `relay-v2` one, as it seems more explicit.

transit.md
```diff
@@ -167,23 +170,164 @@ The Transit client can attempt connections to multiple relays, and uses the
...
+## Canonical abilities encodings
+
+The transit protocol relies on an existing secured (but possibly low-bandwidth)
+communication channel to exchange the abilities and hints. Thus, it has no
+influence over how they are encoded. However, we make a suggestion using JSON
+messages for other protocols to use.
+
+Abilities are encoded as a list, each item having a `type` tag. An ability may
+not appear more than once in the list. Setting `relay-v2` requires also setting
+`relay-v2` for backwards compatibility.
```
meejah, line 188: This one should say `relay-v1` right?

transit.md
```diff
@@ -167,23 +170,164 @@ The Transit client can attempt connections to multiple relays, and uses the
...
+{
+  "type": "direct-tcp-v1" or "tor-tcp-v1",
+  "hostname": <string>,
+  "port": <u16>,
+  "priority": <float; optional>
+}
+```
+
+`hostname` must be compliant with the `host` part of an URL, i.e. it may be an
+IP address or a domain. Furthermore, IPv6 link-local addresses are not supported.
```
meejah, line 226: Definitely like the hostname restriction .. why the caveat about link-local addresses (and why is it v6-only?) Could link-local address be useful during testing? (If they're not, we should exclude IPv4 ones too though, no?)

transit.md
```diff
@@ -167,23 +170,164 @@ The Transit client can attempt connections to multiple relays, and uses the
...
+  "type": "relay-v2",
+  "name": "<string, optional>",
+  "hints": [
+    {
+      "type": "tcp",
+      "hostname": "<string>",
+      "port": "<tcp port>"
+    },
+    {
+      "type": "websockets",
```
meejah, line 255: `websockets` -> `websocket`

transit.md
```diff
@@ -167,23 +170,164 @@ The Transit client can attempt connections to multiple relays, and uses the
...
+      "url": "<url>"
+    },
+    …
+  ],
+}
+```
+
+A relay server may be reachable at multiple different end points and using
+different protocols. All hinted end points for a relay server must point to the
```
meejah, line 264: "end points" -> "endpoints"

transit.md
```diff
@@ -167,23 +170,164 @@ The Transit client can attempt connections to multiple relays, and uses the
...
+same location, and the relay server must be able to connect any two of these
+endpoints. If this is not the case, the relay server must be advertized using
+multiple distinct hints instead (one per endpoint). Furthermore, a relay server
+may be given a human readable name for UI purposes.
+
+Hints have a `type` field, of which the currently known values are `tcp` and
+`websockets`. Hints of unknown type must be ignored. The `type` field is only
```
meejah, line 271: `websockets` -> `websocket`

transit.md
```diff
@@ -167,23 +170,164 @@ The Transit client can attempt connections to multiple relays, and uses the
...
+required for `relay-v2` encoding. In the `relay-v1` incoding, only `tcp` hints
+are supported and the `type` field is implicit.
+
+The encoding of a `tcp` hint is equivalent to the one used by `direct-tcp-v1`.
+A hint of type `websockets` has an `url` field instead, which points to the
+WebSocket. Both `ws://` and `wss://` protocols must be supported.
```
meejah, line 277: Why _must_ unencrypted connections be supported? (I understand that there's encryption already so using `wss` is maybe redundant, but also browsers or other restricted environments may not be allowed to make non-`wss` connections .. so tempted to change this to something like:

> The `wss://` protocol must be supported and we recommend supporting `ws://` as well whenever possible (the underlying messages are already encrypted by default).

transit.md
```diff
@@ -167,23 +170,164 @@ The Transit client can attempt connections to multiple relays, and uses the
...
+  "type": "relay-v2",
+  "name": "Example relay",
+  "hints": [
+    {
+      "type": "tcp",
+      "hostname": "example.org",
+      "port": "1234"
+    },
+    {
+      "type": "websockets",
```
meejah, line 307: websockets -> websocket

meejah (commented):
looking good!
see also https://github.com/magic-wormhole/magic-wormhole-protocols/pull/16#discussion_r792205497

transit.md
piegamesde: Link-local addresses are bound to the used interface and are thus not compatible across machines even within the same local network. Furthermore, I've made the experience that the required interface suffix is not handled well across the stack (parsers, libraries) — not even Chrom*/Firefox support these URLs; both closed the issue as *wontfix* IIRC.

"Normal" local IPv6 addresses are still allowed of course, although the concept is rather weird since most IPv6 addresses are in a global namespace regardless (no NATting).

transit.md
piegamesde: That's a good question, I obviously didn't think it through. The question is, *who* must *what*: we may either put requirements on the client side or on the relay side; the requirements for the others follow from that.

My initial idea was that clients must support both since a relay may not provide `wss` (this does not really hurt since all libraries should support it; it's more a question of policy). Alternatively, we could require all relays to support both. Then, it would not matter which one is used in the URL and it would be up to the client which one to use.

meejah: A client in a Web browser won't be able to support `ws://` in general.

piegamesde: Wait, what? Do they already enforce TLS?

transit.md
meejah: Okay, maybe link-local addresses are useless .. but aren't v4 ones useless too then?

piegamesde: No, since they are scoped per network and not per interface. As soon as two peers are in the same network, it works fine.

transit.md
piegamesde: I thought on this some more, and made two observations:

- Unlike for the rendezvous server, TLS does not change anything to our security model
- Even if it did, hosting a relay server for a private network would be just fine with or without TLS

Thus, I propose the following:

- If the Relay server supports TLS, it should be advertized as `wss`
- `wss` implies `ws`, i.e. insecure WebSockets connections must always be accepted
- The clients are free in their choice, but encouraged to support both.

transit.md
meejah: Okay, I see .. what about 3 Docker containers on the same host, those are still "different interfaces" for IPv6 because each container has its own (virtual) one?

piegamesde: I'm not sure, it probably depends on the setup. In any case, we *always* have "normal" IPv6 addresses: we need internet access (or at least a network) to access the rendezvous server.

transit.md
meejah: For Web browsers: it _might_ be possible to think of situations where you'd be able to do unencrypted connections, but the page itself should be served by TLS which as far as I understand means you can't make unencrypted WebSocket connections. (I'm not 100% sure about this, browsers are fickle beasts ;) but [MDN](https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_WebSocket_client_applications) says, "This should use the URL scheme wss://, although some software may allow you to use the insecure ws:// for local connections. "

I guess partially I'm just reluctant to say "you _must_ support unencrypted connections" these days: that would mean someone deploying this has to understand enough about the protocols etc to learn that the underlying data packets are ("probably" -- that seems to be optional from the wording?) encrypted already. I'm presuming devops practice these days is normalizing on "everything is transport-encrypted" -- that's why I wanted to leave room in the spec. wording for someone to choose "everything is encrypted, always, and we don't allow unencrypted protocols".

(That is, an implementer could easily conclude from "must support both" that even having an option to turn one or the other off is bad).

piegamesde: Oh I didn't know this, but it makes sense. How about this then:

- Both relay servers and clients *should* support `wss`, and it should be preferred over `ws`.
- If a relay server supports both, advertise that endpoint using two hints.

meejah: yeah that sounds good

piegamesde:
All comments are resolved, maybe we all read through it once or twice to catch even the last typo and then it should be good to go :tada:

transit.md
```diff
@@ -167,23 +170,167 @@ The Transit client can attempt connections to multiple relays, and uses the
...
+    …
+  ],
+}
+```
+
+```json
+{
+  "type": "relay-v2",
+  "name": "<string, optional>",
+  "hints": [
```
piegamesde, line 248: I thought I'd implement the proposed changes before merging (https://github.com/magic-wormhole/magic-wormhole.rs/pull/138/commits/d2771f28ae8a0293d26fe5d6f7f629b79e329a88). During this, I got a bit confused about the naming.

I think renaming `hints` to `endpoints` or something would make things less confusing.

piegamesde: Okay I tried it out in the code, and it was a marginal improvement. However, while trying to writing it in the protocol I found that it would make the text more complicated. So in the end, it's probably not worth it.

meejah:
awesome (and sorry this took so long to get merged)

PR #17, opened 2022-03-20
Add session resumption through seeds

### LINKED ISSUE #2: Wormhole Seeds
Migrated from https://github.com/magic-wormhole/magic-wormhole/issues/77, as I think it's a good idea to discuss feature that are not Python-exclusive here.

-------

Basically, both sides of a Wormhole connection would derive a 128-bit mailbox-id and a 256-bit wormhole secret (from the PAKE session key), and store it for later use as a "Seed". This seed is used exactly like a normal wormhole code, except that the mailbox ID is used directly (instead of being treated as a "nameplate" which then points to a mailbox), and the Seed can be reused.

(we don't strictly need to use PAKE each time, but it happens to provide forward-secrecy, and we already have all the code in place.. it'd actually be more work to use a simple non-PAKE KDF).

(also, the wormhole secret could be considerably shorter, and still be safe, but there's no harm in making it full-sized)

piegamesde:
Closes #2.

I integrated part of the feedback by focusing more on the client protocol API. Implementation notes have now been moved to a separate sub-section.


client-protocol.md
```diff
@@ -103,3 +103,77 @@ will be released, and the WebSocket connection will be dropped.
...
+# `derive(key, purpose)` is the usual key derivation function
+seed = hex(derive(session_key, "wormhole:seed"))
+```
+
+The `seed` is the main shared secret between the peers and all other data will
+be derived from it:
+
+```python
+password = hex(derive(seed, "wormhole:seed:password"))
+nameplate = hex(derive(seed, "wormhole:seed:nameplate"))
```
piegamesde, line 153: Using a high entropy nameplate instead of connecting to a mailbox directly solves a lot of the connection issues that I previously had. Notably, multiple concurrent independent connections are possible using the same seed because they will end up in a different mailbox. Also, for the same reason, the error recovery when a mailbox gets "crowded" is a lot quicker.

The current rendezvous server implementation supports this without issues, but nevertheless it may be a good idea to codify this in the protocol that we now rely on this possibility.

PR #19, opened 2022-03-24
Transit: WebSocket relays, take II

piegamesde:
This reverts the last commit of #16 and replaces it with a new alternative. I already checked that the current formulation is compatible with existing clients, but somebody more familiar with the code bases please double check that.

### DIFF
--- a/file-transfer-protocol.md
+++ b/file-transfer-protocol.md
@@ -76,16 +76,32 @@ The `transit` message mentioned above is encoded following this schema:
 ```json
 {
     "transit": {
-        "abilities-v1": [ … ],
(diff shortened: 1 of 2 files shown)


meejah (approved):
LGTM

(After reading that, I wonder if we want `websocket-v1` or similar instead of just `websocket`...?)

piegamesde:
> I wonder if we want websocket-v1 or similar

I don't have strong opinions either way. I think that the others should not have been versioned instead, but now it's too late so maybe it would be better to at least keep it consistent?

-------

CC @psanford, please confirm that this would be compliant with what the Go implementation currently does.

PR #21, opened 2022-05-14
File transfer URI Scheme

### LINKED ISSUE #3: Standardize on a QR Code format?
Recently I've been working on adding QR Code support to the wormhole-william cli[1] and Android[2] apps. The idea is if you want to do a transfer from your computer to your phone, you can generate a QR code to scan instead of having to type the code into your phone.

I don't know if this is of interest to any other Magic Wormhole implementations. If it is, it would be nice to share a common url format embedded into the QR Code. I propose the following format:

```
wormhole:${relay_url}?code=${code}
```

So for example, using the default relay with a code of ` 1-babylon-locale` the embedded code would be:

```
wormhole:ws://relay.magic-wormhole.io:4000/v1?code=1-babylon-locale
```

If there is interest, I can open a PR to document this.

[1]: https://github.com/psanford/wormhole-william/pull/37
[2]: https://github.com/psanford/wormhole-william-mobile/pull/7

felinira:
Add a `wormhole-transfer` URI scheme.

Closes #3

### DIFF
--- a/uri-scheme.md
+++ b/uri-scheme.md
@@ -1,0 +1,43 @@
+# URI Scheme
+## Scope
+This URI scheme describes the wormhole file transfer application. For other use cases of the wormhole spec and URI scheme has not been defined yet.
+
+The application ID for this protocol is always assumed to be `lothar.com/wormhole/text-or-file-xfer`
+
+## Description
+
+The `wormhole-transfer` URI Scheme is used to encode a wormhole code for file transfer as a URI. This can then be used to generate QR codes, or be opened by the platform URI handler to open a supporting client.
+
+The general format looks like this, and assumes default values for all query fields:
+`wormhole-transfer:{code}`
+
+`{code}` is the [URL / percent encoded](https://en.wikipedia.org/wiki/Percent-encoding) wormhole code. While common codes may not require any encoding, it must be made sure that percent-encoding and decoding is applied to support all possibilities.
+
+It can be extended by appending a [query string](https://en.wikipedia.org/wiki/Query_string).
+
+## Query fields
+Applications MUST parse all query fields specified below and fail if they contain unknown fields or unsupported parameter values. Query values are percent-encoded.
+
+* `version`: The version of the URI scheme.  At the moment, only version 0 is specified. Clients must check this for compatibility.
+Default: `0`
+
+* `rendezvous`: The rendezvous server and protocol to use, including its port. The versioned endpoint is supposed to be added by the client implementation, eg. `/v1`
+Default: `ws://relay.magic-wormhole.io:4000`
+
+* `role`: The type of operation requested. Valid values are:
+  * `follower`: The URI *parsing* client is supposed to be prepared to receive a file over the connection once established.
+  * `leader`: The URI *parsing* client is supposed to send a file. This functionality can be used if it is easier (from the user's point of view) to read the URI in the opposite direction (for example because QR codes are used and only one device is equipped with a camera).
+Default: `follower`
+
+## Examples
+To encode the code `4-hurricane-equipment` to send a file the URI would look like this:
+
+`wormhole-transfer:4-hurricane-equipment`
+
+By expanding the default values this would be equivalent to:
+
+`wormhole-transfer:4-hurricane-equipment?version=0&rendezvous=ws%3A%2F%2Frelay.magic-wormhole.io%3A4000&role=follower`
+
+To request the URI parsing side to lead the connection instead (and probably send a file), the URI would look like this:
+
+`wormhole-transfer:4-hurricane-equipment?role=leader`
--- a/index.rst
+++ b/index.rst
@@ -19,6 +19,7 @@ Magic-Wormhole: Get Things From One Computer To Another, Safely
    server-protocol
    client-protocol
    file-transfer-protocol
+   uri-scheme
 
    attacks
    journal


uri-scheme.md
```diff
@@ -0,0 +1,35 @@
...
+It can be extended by appending a [query string](https://en.wikipedia.org/wiki/Query_string).
+
+## Query fields
+Applications MUST parse all query fields specified below and fail if they contain unknown fields or unsupported parameter values. Query values are [URL encoded](https://en.wikipedia.org/wiki/Percent-encoding).
+
+* `version`: The version of the URI scheme.  No other value than `0` is accepted.
+Default: 0
+
+* `rendezvous`: The rendezvous server and protocol to use, including its port.
+Default: `ws://relay.magic-wormhole.io:4000/v1`
```
piegamesde, line 17: The `/v1` must not be part of the URL, and instead must be added by the client implementation. The assumption is that a rendezvous server may have different versioned endpoints that are all compatible with each other.

uri-scheme.md
```diff
@@ -0,0 +1,35 @@
+# URI Scheme
+
+The `wormhole` URI Scheme is used to encode a wormhole code as a URI. This can then be used to generate QR codes, or be opened by the platform URI handler to open a supporting client.
+
+The general format looks like this, and assumes default values for all query fields:
+`wormhole:{code}`
```
piegamesde, line 6: Since codes may more or less be arbitrary, maybe add a note on escaping here too

uri-scheme.md
piegamesde, line 17: The `/v1` must not be part of the URL, and instead must be added by the client implementation. The assumption is that a rendezvous server may have different versioned endpoints that are all compatible with each other.

uri-scheme.md
```diff
@@ -0,0 +1,35 @@
...
+## Query fields
+Applications MUST parse all query fields specified below and fail if they contain unknown fields or unsupported parameter values. Query values are [URL encoded](https://en.wikipedia.org/wiki/Percent-encoding).
+
+* `version`: The version of the URI scheme.  No other value than `0` is accepted.
+Default: 0
+
+* `rendezvous`: The rendezvous server and protocol to use, including its port.
+Default: `ws://relay.magic-wormhole.io:4000/v1`
+
+* `type`: The type of operation requested. Valid values are:
```
piegamesde, line 19: What would be the use case for this? Normally the sender generates the code and the receiver enters it, why should it be any different when encoding it in a URI?

uri-scheme.md
```diff
@@ -0,0 +1,35 @@
...
+		* `recv`: The URI generating client is asking to send a file and is requesting the URI parsing client to receive a file.
+		* `send`: The URI generating client is asking to receive a file and is requesting the URI parsing client to offer to send a file.
+Default: `recv`
+
```
piegamesde, line 23: Please add a parameter for the [AppID](https://github.com/magic-wormhole/magic-wormhole-protocols/blob/114ebb88ec6c6d34711d94beb0257b1d6b4cd0f6/server-protocol.md#application-ids=), making it default to file transfer

uri-scheme.md
```diff
@@ -0,0 +1,35 @@
...
+The general format looks like this, and assumes default values for all query fields:
+`wormhole:{code}`
+
+It can be extended by appending a [query string](https://en.wikipedia.org/wiki/Query_string).
+
+## Query fields
+Applications MUST parse all query fields specified below and fail if they contain unknown fields or unsupported parameter values. Query values are [URL encoded](https://en.wikipedia.org/wiki/Percent-encoding).
+
+* `version`: The version of the URI scheme.  No other value than `0` is accepted.
```
piegamesde, line 13:
```suggestion
* `version`: The version of the URI scheme. At the moment, only version 0 is specified. Clients must check this for compatibility.
```

piegamesde (changes requested):
Looks pretty solid to me

uri-scheme.md
felinira: Not all devices have a camera. It might be convenient to generate the code on a computer for example, scan it from a phone, select the file and send it right away.

piegamesde:
How about using `magic:` as scheme instead of `wormhole:`? :D

uri-scheme.md
piegamesde: Interesting. I'm not opposed, but I'd like to discuss this a bit more since it is the only feature that I wouldn't consider as trivial.

To reiterate: The URL is generated by the sender and the receiver scans it. This requires the receiver to have a camera. Otherwise, the code must be transferred as text, either by typing it in old style or by sending over the link (you could also send the code without link, either way). Your proposal here is that when the receiver does not have a camera (e.g. phone -> desktop transfer), the code is generated on receiver side and the sender then can scan it.

Protocol wise this should be fine and easy enough to implement with most libraries. But how do you imagine the UI side, both for CLI and GUI?

felinira: For my application I would probably have a QR button next to the code entry on the receive tab which would generate a code that could be scanned by the other side. After scanning the code the sending application would understand that this is actually a request and display a message with a button to open up a file picker. After picking a file the file would be sent automatically.

I attached a few mockup screenshots, I would probably change a few things up, but you get the idea.

It is definitely more work to implement for everyone, and might also create a small amount of confusion. But in my opinion this would make the very common workflow of "I have something on my phone and I need it on my computer" way easier.

Requesting side:
image
image

Sending side:
image

felinira: For CLI I think the only thing to do would be to add a `--request` flag to `wormhole receive` that would generate this code, and then wait for connection and save the file

I don't know how the CLI would want to support the URI scheme on the parsing side. It might allow URI entry for the code, which would prompt to enter a file name if the code direction would be reversed. This would not be the best workflow, but I also don't see any reason why someone would want to do this with the CLI apart from doing it by accident.

meejah: Maybe haven't thought this _completely_ through, but it could also work if the CLI did `send` along with a "request" URI -- that is, semantically you're still "sending" from the CLI side.

The Python CLI already has some interactive modes, so it _could_ prompt for a filename (although I agree this is a little weird for CLIs and doing e.g. tab-completion would probably be "expected" but also non-trivial).

So basically something like `wormhole send --uri <valid request URI> /path/to/local/file/to/send`

uri-scheme.md
```diff
@@ -0,0 +1,40 @@
...
+
+* `type`: The type of operation requested. Valid values are:
+		* `recv`: The URI generating client is asking to send a file and is requesting the URI parsing client to receive a file.
+		* `send`: The URI generating client is asking to receive a file and is requesting the URI parsing client to offer to send a file.
+Default: `recv`
+
+## Examples
+To encode the code `4-hurricane-equipment` to send a file the URI would look like this:
+
+`wormhole:4-hurricane-equipment`
```
meejah, line 32: :+1:

meejah:
Something to think about is "generic-ness" (maybe there's a better word) .. Although the main use-case for "magic wormhole" is the file-transfer application, in principal it's useful to consider two layers: the core of the scheme is "a way to set up a secure communications (via one-time code)" and the "application" layer can do whatever it likes with that -- with "file transfer" being widely implemented.

I'm not sure if this points to anything needing changes in this proposal (except _possibly_ re-naming the `type` field or its values to e.g. `file-send` and `file-recv`) but the reason I bring it up is to consider what happens if a new widely-used application exists that we'd want to add to this spec. (As a thought-experiment, that could be "exchange SSH identities" since that's been suggested before and I believe the Python implementation had/has code for that).

It seems like that could totally fit with the existing scheme simply by adding `type=ssh` or similar.

Perhaps that suggests re-naming `?type=` to `?app=` or similar -- making its use as "a way to specify an application protocol to speak" more explicit.

If we agree on that, then it would likely make sense to include words like "implementations SHOULD accept unknown values for app= when parsing the URI".

piegamesde:
I think what you propose is already covered by the `appid` parameter which I requested in the first round of review. However I agree that `type` may not be the best name for that parameter.

But your concerns about different application types raises a new question: Should we have one URI scheme per AppID? Because let's say that I have an application that does port forwarding instead of file transfer. Which app should now associate with the `wormhole:` URI scheme?

meejah:
> I think what you propose is already covered by the `appid` parameter which I requested in the first round of review. However I agree that `type` may not be the best name for that parameter.
> 
> But your concerns about different application types raises a new question: Should we have one URI scheme per AppID? Because let's say that I have an application that does port forwarding instead of file transfer. Which app should now associate with the `wormhole:` URI scheme?

Yeah, I was going to write something about `?appid=` too ... there's certainly some interaction there. I was considering the other way around as well: a particular `?app=` (or `?type=`) is associated in the spec with a particular `appid`, approximately. But, I also like the open-ended-ness of having `appid` in there explicitly .... 

I guess overall, we'd have to also allow open-ended query params for `?appid=` to be generally useful since each app will no doubt have some sorts of options etc to convey in the URI.

Perhaps combining these leads to:
- `?app=` is strongly associated with a particular app-id, and also lists valid (additional) query-args in this document
- default `?app=` is `filetransfer` or so (with associated defaults for appid and `?type=` or whatever it should be called there)
- you cannot say both `?app=` and `?appid=` in the same URI
- parsers should accept any additional query-args at all for unknown `?app=` or `?appid=` URIs (but can be more strict for known `?app=` names that are defined in this document).

This means that implementations can experiment with new apps (including new URI query-args) and once they are "well-known" enough (or specified-enough) a PR to this repository can add them to this document as a well-defined `?app=` name.

uri-scheme.md
piegamesde: I think for the Rust implementation, I'll go with something like `wormhole receive-request` which gives you a code. Then, the sender will hat to do `wormhole send --code`, which is the same thing as using custom code which is supported already. It's important to me that the URIs behave identically to codes, so at any point in time one may choose to use the one or the other.

That being said, let's pain that shed: For transit, we use "leader" and "follower" roles instead of "sender" and "receiver" to make more clear it is protocol agnostic. Also "type" may be a bit confusing (see @meejah's comment below), so how about "role" or "direction"?

felinira:
> But your concerns about different application types raises a new question: Should we have one URI scheme per AppID? Because let's say that I have an application that does port forwarding instead of file transfer. Which app should now associate with the `wormhole:` URI scheme?

This is a difficult one. I probably agree. There might be multiple applications and it might not be desirable to support every protocol extension with every app.

`wormhole-file` and `wormhole-ssh` are a bit longer but might be more flexible in the long term.

I don't know if we need to expand the URI scheme to all apps and all possibilities. If an app wants to do something special that only itself would understand they are free to define their own URI scheme after all. Changing appid is fine I guess (even though I fail to see many reasons for it except for maybe some sort of rate limiting), but behaving out of spec should not be encouraged by the spec.

uri-scheme.md
felinira: `role=leader` means the URI parsing application would have to ... do something? Send a file? Decide what it wants to do instead of executing something predetermined? Not sure yet what it would mean in practice.

What about a potential new mode that would allow for file transfer in both directions? Who would be who in that case? I don't propose we add that here right away, just wondering how it could look like.

felinira:
> Another option might be like `wormhole:file:....` (and `wormhole:ssh:...` for a hypothetical ssh app protocol).

I don't think this is compatible with URL parsers (so far the spec has been and this has some advantages) and you definitely can't use this as a URI scheme in a desktop application (like in a freedesktop spec desktop file with `MimeType=x-scheme-handler/wormhole`). The idea of `wormhole-file:` would be that different applications would be able to register different protocol prefixes, according to what they support. This is not possible with this proposal. If we don't want that, I'd prefer using a regular query parameter instead.

> What we could do for _this_ PR is define a URI scheme for just the file-transfer application since that protocol a) exists and b) has several implementations (e.g. `wormhole-file:...` or so). That would still allow for a more-general `wormhole:....` scheme to be defined and direct future (application-layer) protocol designers to put all their options etc into mailbox messages (if we agree this would work). Thus, if there were a "version 2" file-transfer protocol it could use the "general" scheme.

I wouldn't mind scoping this to file transfer, actually I think this is probably a great idea. I can modify the scope of the PR if nobody has objections to that. Extending from there can happen whenever someone has a better idea.

meejah:
I'm +1 on scoping this to `wormhole-file:` or similar; we can spec out a more-general version later. That takes out a bunch of complexity and doesn't force us to thing about those more-complex issues (yet!)

psanford:
I'd appreciate an attempt to keep compatibility with what we're already using in wormhole-william{,-mobile}.

Specifying the app-id in the qr-code doesn't really make much sense to me. The app-id is just used to namespace the mailbox server. Its not really part of the wormhole protocol, its part of the rendezvous protocol (i.e. if you wanted to build a non-wormhole app that used a rendezvous server for coordination you'd use a different app-id). 

The `file` vs `ssh` distinction doesn't make sense to me. The type of transfer is communicated between the two clients post handshake, there's no reason to include it in the qr code (just like it isn't included in the text wormhole code when you use the cli today).

meejah:
All wormhole connections are via the rendezvous server, and you need an app-id to make that connection.

meejah:
> Yes, but if you are going to do a wormhole handshake your app id will be lothar.com/wormhole/text-or-file-xfer. Using wormhole: implies that app-id.

okay I think that's where our misunderstanding is here; "wormhole" is the generic setup stuff and something like "wormhole-file" is "the generic setup stuff _and_ the particular application protocol to speak (namely: file-transfer)".

piegamesde:
I think we have the following options:

- Have a file transfer specific URI scheme (like `wormhole-file`, `wormhole-transfer` or `w-transfer`).
  - In that case no `appID` is required and it may be hardcoded to `lothar.com/wormhole/text-or-file-xfer'
  - Other app IDs would have to register their own scheme, although the content would probably look really similar
- Have one URI scheme for all Wormhole related things, with the necessity to specify an app ID (optional via default value) and the previously discussed downsides
- Have both??

Some random notes on stuff said above:

- Due to the way the protocols are stacked this won't be fixable in Transfer v2.
- As pointed out by @felinira, `wormhole:file:` sadly won't work as we'd like it to
- Depending on how it is currently implemented, having a separate scheme for `ssh` does not make that much sense; this should be handled by the file transfer protocol. A better example to look at would be the "file transfer" vs "port forwarding" use case.

meejah:
> Depending on how it is currently implemented, having a separate scheme for ssh does not make that much sense; this should be handled by the file transfer protocol. A better example to look at would be the "file transfer" vs "port forwarding" use case.

Yeah, ssh was "just a thought-experiment": I had in mind something like magic-folder invites, both of which are not like file-transfer because the amount of data transferred is "not much" so there's no reason to set up an actual transit connection; the key-exchange is just via another message (or two or three) in the mailbox. In the "invites" case there's key material going both ways. In any case, either one serves as an example of "some app protocol which is not the existing file-transfer protocol" (even if one could imagine making the file-transfer protocol serve for a lot of use-cases _like_ these).

> Have both??

Yeah, I think this might be the longer-term approach -- with the advantage that we can just spec out the much narrow (and hence easier) existing file-transfer use-case by itself, right now. Could be `magic-file:` or so as well, to capture the `magic:` idea from a previous comment.

Agree with the other notes.

felinira:
I will rewrite this PR to be about the file-transfer URI only then.

Please react to this message with your preferred protocol name(s) for file transfer.
:rocket: `wormhole-file:`
:smile: `wormhole-transfer`
:tada: `magic-file:`
:eyes: `w-transfer`
other: please comment

psanford:
What the reason to not use what we've already implemented and have been using for over a year in wormhole-william, specifically what was discussed in #3 ?

piegamesde:
Because, IMO, it has the arguments swapped. Why is the rendezvous server always specified although it is technically optional and the code an argument even though it will always be present?

Also, simply putting an URL as the path of another might lead to interesting edge cases without proper encoding. For example, what if a query parameter is part of the rendezvous server URL?

meejah:
Agree with the above comment.
This makes the "use all defaults" URI look like `wormhole-transfer:1-code-words` or so (depending on scheme name) which is nice and compact. It should be the "rare" case where you need to specify the rendezvous server location (and relay server hints are already in-protocol).

uri-scheme.md
```diff
@@ -0,0 +1,43 @@
...
+Applications MUST parse all query fields specified below and fail if they contain unknown fields or unsupported parameter values. Query values are percent-encoded.
+
+* `version`: The version of the URI scheme.  At the moment, only version 0 is specified. Clients must check this for compatibility.
+Default: `0`
+
+* `rendezvous`: The rendezvous server and protocol to use, including its port. The versioned endpoint is supposed to be added by the client implementation, eg. `/v1`
+Default: `ws://relay.magic-wormhole.io:4000`
+
+* `role`: The type of operation requested. Valid values are:
+		* `follower`: The URI *generating* client is asking to lead the transit connection (and most likely send a file) and is requesting the URI *parsing* client to follow the leader (and most likely receive a file).
```
piegamesde, line 28: Style nit: other parts of the documentation use 2 or 4 spaces for indentation

felinira: Whoops, that was a mistake :D

uri-scheme.md
```diff
@@ -15,26 +21,23 @@ Applications MUST parse all query fields specified below and fail if they contai
...
 ## Examples
 To encode the code `4-hurricane-equipment` to send a file the URI would look like this:
 
-`wormhole:4-hurricane-equipment`
+`wormhole-transfer:4-hurricane-equipment`
 
 By expanding the default values this would be equivalent to:
 
-`wormhole:4-hurricane-equipment?version=0&appid=lothar.com%2Fwormhole%2Ftext-or-file-xfer&rendezvous=ws%3A%2F%2Frelay.magic-wormhole.io%3A4000%2Fv1&type=recv`
+`wormhole-transfer:4-hurricane-equipment?version=0&appid=lothar.com%2Fwormhole%2Ftext-or-file-xfer&rendezvous=ws%3A%2F%2Frelay.magic-wormhole.io%3A4000%2Fv1&role=leader`
```
meejah, line 39: +1, like "role" better (and matches other langauge in magic-wormhole)

felinira: I forgot to remove appid from this URI though

uri-scheme.md
```diff
@@ -0,0 +1,43 @@
...
+Default: `0`
+
+* `rendezvous`: The rendezvous server and protocol to use, including its port. The versioned endpoint is supposed to be added by the client implementation, eg. `/v1`
+Default: `ws://relay.magic-wormhole.io:4000`
+
+* `role`: The type of operation requested. Valid values are:
+  * `follower`: The URI *generating* client is asking to lead the transit connection (and most likely send a file) and is requesting the URI *parsing* client to follow the leader (and most likely receive a file).
+  * `leader`: The URI *parsing* client is supposed to lead the transit connection (and most likely send a file) and the URI *generating* client is asking to follow the leader (and most likely receive a file). This functionality can be used if it is easier (from the user's point of view) to read the URI in the opposite direction (for example because QR codes are used and only one device is equipped with a camera). Clients with URI support are required to support this functionality on the parsing side, but must not necessarily allow to generate such a URI itself.
```
felinira, line 29: I think we need to specify what 
> Clients with URI support are required to support this functionality on the parsing side, but must not necessarily allow to generate such a URI itself.

means. Should they be required to support this mode or only be required to understand what this means? I am leaning to the latter.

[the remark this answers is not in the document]
meejah: I might choose wording like, "Clients with URI support MUST support parsing leader-mode URIs and MAY support generation of appropriate leader-mode URIs" but not sure if we're really following RFC2119 generally...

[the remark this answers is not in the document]
piegamesde: Honestly, I think that sentence could just be removed. It is written elsewhere that parsers must look at all arguments, and generators will only implement as much features as they need anyways.

piegamesde (changes requested):

uri-scheme.md
```diff
@@ -0,0 +1,43 @@
...
+## Query fields
+Applications MUST parse all query fields specified below and fail if they contain unknown fields or unsupported parameter values. Query values are percent-encoded.
+
+* `version`: The version of the URI scheme.  At the moment, only version 0 is specified. Clients must check this for compatibility.
+Default: `0`
+
+* `rendezvous`: The rendezvous server and protocol to use, including its port. The versioned endpoint is supposed to be added by the client implementation, eg. `/v1`
+Default: `ws://relay.magic-wormhole.io:4000`
+
+* `role`: The type of operation requested. Valid values are:
```
piegamesde, line 27: I got some feedback that this section is written in a confusing way. I think one step would be to phrase it specifically to the file transfer use case (but keeping the `leader` and `follower` values in order to better recycle the parser in the future).

felinira:
I finally got around to fixing the wording here. FCP?

uri-scheme.md
```diff
@@ -0,0 +1,43 @@
...
+Default: `follower`
+
+## Examples
+To encode the code `4-hurricane-equipment` to send a file the URI would look like this:
+
+`wormhole-transfer:4-hurricane-equipment`
+
+By expanding the default values this would be equivalent to:
+
+`wormhole-transfer:4-hurricane-equipment?version=0&rendezvous=ws%3A%2F%2Frelay.magic-wormhole.io%3A4000%2Fv1&role=follower`
```
piegamesde, line 39:
```suggestion
`wormhole-transfer:4-hurricane-equipment?version=0&rendezvous=ws%3A%2F%2Frelay.magic-wormhole.io%3A4000&role=follower`
```

piegamesde:
One last nit commented above. Please make sure that the formatting is consistent (mostly whitespace and newlines) and make the git history pretty; otherwise it's good to go for FCP.

felinira:
> One last nit commented above. Please make sure that the formatting is consistent (mostly whitespace and newlines) and make the git history pretty; otherwise it's good to go for FCP.

done

piegamesde:
:bell: Let's start the Final Comment Period then :bell:

Since this has already been reviewed rather thoroughly, a week should suffice.

uri-scheme.md
```diff
@@ -0,0 +1,43 @@
+# URI Scheme
+## Scope
+This URI scheme describes the wormhole file transfer application. For other use cases of the wormhole spec and URI scheme has not been defined yet.
```
meejah, line 3: These two read a little awkwardly to me.

What about, "This URI scheme pertains to the wormhole file-transfer application. A URI specification covering other use-cases of magic-wormhole has not yet been defined".

uri-scheme.md
```diff
@@ -0,0 +1,43 @@
...
+
+It can be extended by appending a [query string](https://en.wikipedia.org/wiki/Query_string).
+
+## Query fields
+Applications MUST parse all query fields specified below and fail if they contain unknown fields or unsupported parameter values. Query values are percent-encoded.
+
+* `version`: The version of the URI scheme.  At the moment, only version 0 is specified. Clients must check this for compatibility.
+Default: `0`
+
+* `rendezvous`: The rendezvous server and protocol to use, including its port. The versioned endpoint is supposed to be added by the client implementation, eg. `/v1`
```
meejah, line 24: Replace "The rendezvous server and protocol..." with "The base URI of the rendezvous server, including protocol and port, to use."

meejah (commented):
Some wording nit-picks inline.

Since we use "MUST" at least once here, might be worth brief mention of RFC2119?

piegamesde:
I don't think we use RFC2119 elsewhere, so maybe rather change SHOULD to an informal should to keep it consistent with the rest?

felinira:
I applied the above suggestions to the wordings.

> I don't think we use RFC2119 elsewhere, so maybe rather change SHOULD to an informal should to keep it consistent with the rest?

I changed MUST to "are required to". If it seems useful to use RFC2119 there should probably be a PR that adds it everywhere. Only using it once does not really justify mentioning it here.

felinira:
I *think* the week is over now^^.

piegamesde:
Yeah, some weeks are just longer than others ;)

PR #26, opened 2022-08-10
A few fixes to make the protocol spec compatible with the code

vu3rdd:
`can-dilate` in the versions message contains the versions list. The `please` message no longer has the `use-versions` key.

Also `please-dilate` is called `please` in the code.

### DIFF
--- a/dilation-protocol.md
+++ b/dilation-protocol.md
@@ -24,21 +24,21 @@ version present in both side's lists is eligible for use.
 ## Leaders and Followers
 
 Each side of a Wormhole has a randomly-generated dilation `side` string (this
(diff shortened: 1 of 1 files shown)


piegamesde:
I agree with the change. But I think it might be a good idea to collect all of these Dilation-related changes in a dedicated branch. (FWIW, this change is already contained in the `dilation-reword` branch)

meejah:
I think the `use-version` should probably be put into the code .. it's essentially "just" a confirmation that the Follower has picked the correct version (but also being explicit is good).

piegamesde:
Please retarget the PR against the `dilation` branch, I'll merge it then

PR #27, opened 2022-09-03
Clarify percent-encoding used in URI spec

felinira:
Clarify the use of percent encoding used in the host and query parts of the URI.

This clarification ensures full compatibility with the [URL standard](https://url.spec.whatwg.org/). Any URL that is not one of the "special" ones (ftp, file, http, https, ws, wss) are supposed to use percent encoding with the C0 character set for the host and path components.

### DIFF
--- a/uri-scheme.md
+++ b/uri-scheme.md
@@ -11,9 +11,9 @@ The `wormhole-transfer` URI Scheme is used to encode a wormhole code for file tr
 The general format looks like this, and assumes default values for all query fields:
 `wormhole-transfer:{code}`
 
-`{code}` is the [URL / percent encoded](https://en.wikipedia.org/wiki/Percent-encoding) wormhole code. While common codes may not require any encoding, it must be made sure that percent-encoding and decoding is applied to support all possibilities.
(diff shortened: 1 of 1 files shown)


uri-scheme.md
```diff
@@ -11,9 +11,9 @@ The `wormhole-transfer` URI Scheme is used to encode a wormhole code for file tr
 The general format looks like this, and assumes default values for all query fields:
 `wormhole-transfer:{code}`
 
-`{code}` is the [URL / percent encoded](https://en.wikipedia.org/wiki/Percent-encoding) wormhole code. While common codes may not require any encoding, it must be made sure that percent-encoding and decoding is applied to support all possibilities.
+`{code}` is the [URL / percent encoded](https://en.wikipedia.org/wiki/Percent-encoding) wormhole code. The [C0 percent encode set](https://url.spec.whatwg.org/#c0-control-percent-encode-set) is used to be [compatible with URL parsers](https://url.spec.whatwg.org/#concept-opaque-host-parser). While common codes may not require any encoding, it must be made sure that percent-encoding and decoding is applied to support all possibilities.
```
meejah, line 14: Maybe the text wants to be "C0 control percent encode set" to match the URL and bold text in the standard.

meejah (approved):
LGTM
(...and yeah, we want to be compatible with "URL parsers" as that's sort of the point here, I assume ;)

uri-scheme.md
felinira: Yeah, I must have missed a word there. Changed it :)

meejah:
This is relatively minor and completely in line with prior discussion so I think merging this is fine .. but if others have comments or further clarifications please re-open and/or comment.

PR #32, opened 2023-03-04
Refactoring and improvements

piegamesde:
See commit log. Generally speaking, this changes nothing on the protocol itself. Just the documentation.


security.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Metadata leaks
+
+As usual in cryptography, communication metadata cannot easily be protected. Any
+involved server will know the timestamp and size of each message it relays. This
+usually is not much of an issue, except for file transfer where the file size may
+be characteristic for its content. (Imagine people using Magic Wormhole to send
+some specific movie.)
+
+Thanks to the separation between rendezvous server and relay server, forcing the
```
meejah, line 75: Other places I believe we'd agreed that "mailbox" and "transit" are less-confusing names; perhaps we should use those here now too.

security.md
```diff
@@ -0,0 +1,94 @@
...
+Currently the best defense against this is to use high-entropy nameplates which
+cannot easily be enumerated at the cost of longer codes. This has not be
+implemented yet though, and would still leave the attack vector of "ordinary"
+DoS.
+
+## Metadata leaks
+
+As usual in cryptography, communication metadata cannot easily be protected. Any
+involved server will know the timestamp and size of each message it relays. This
+usually is not much of an issue, except for file transfer where the file size may
```
meejah, line 71: Well, _really_ depends on perspective? (What's a case where it isn't an issue?)

security.md
```diff
@@ -0,0 +1,94 @@
...
+including the IP address of any two peers.
+
+Additionally, in order to establish a direct connection, public and local IP
+addresses are sent to the peer. This may leak information about the local
+network topology, including potentially used VPNs. If you do not trust your
+peer with such information, enforcing a relay server connection is advised.
+However, keep in mind that it is easy to host a relay server, so it might still
+leak your public IP address.
+
+If you do not trust the servers with knowing your IP address either, use
```
meejah, line 91: We generally try to avoid this word .. better is "expose"; that is something like "if you do not want to expose your IP to ..."

security.md
```diff
@@ -0,0 +1,94 @@
...
+[Tor](https://torproject.org/) to use Magic Wormhole anonymously. The Transit
+protocol has Tor support built-in, which then makes use of the Tor network itself
+as the relay.
```
meejah, line 94: I'm not clear what you mean here? The transit is still involved, because we lack support for Onion services (this part is mentioned in a file you've deleted -- perhaps just including that stuff here would be better).

meejah (changes requested):
A lot of this seems straightforward, but I think "security" doesn't accurately re-create the scenarios described in "attacks", and I think we lose a little by deleting the "approaches considered but not tried" (see inline comments).

security.md
```diff
@@ -0,0 +1,94 @@
...
+## Code interception attacks
+
+By default, Wormhole codes contain 16 bits of entropy. Failed attempts of
+guessing a code will cause both clients to error out, thus an attacker
+has a one-in-65536 chance of successfully guessing the code. In the case of
+file transfer, the attacker would then be able to receive the file. As the
+receiver's attempt will fail, such an attack will be noticeable.
+
+If the attacker *additionally* has the capability of intercepting and changing
+network traffic ("MitM"), they might be able to fake arbitrary interactions
```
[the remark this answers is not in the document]
piegamesde: Oh, I see. I interpreted the wording wrong then, will update

security.md
piegamesde: Most of the time actually, the file size does not contain that much information: an attacker gains nothing of value from the fact that my configuration file has 45219 bytes (ignoring for now that the protocol may round numbers up to some block size for encryption).

security.md
piegamesde: Which word?

security.md
piegamesde: I was referring to the Tor part which should be documented in the Transit protocol specification. I'll make it more explicit

piegamesde:
Thanks for the review. Ideally, none of the important information should be lost in the rewrite, so I'll go and add the missing bits

security.md
meejah: File-sizes are actually extremely unique (unfortunately). This e.g. forms part of the basis of "fingerprinting" Web pages, which is extremely successful (e.g. allowing one to answer the question "what website is this TLS connection visiting" even on networks like Tor).

So, I would assume this can be extended to pattern-match on message flows in wormhole as well (allowing fingerprinting of any protocol). I guess it's the "not much of an issue" comment that I'm commenting on here...

security.md
meejah: Ah, sorry ;) "trust"

security.md
meejah: It's still using the transit server though, so the part about "use the Tor network itself as the relay" isn't really correct (or I don't understand what you mean here).

security.md
meejah: Trying to be more-clear: leaking metadata is bad.

So, it's still bad for almost any "application protocol" I can think of, and it should remain clear that Wormhole does NOT protect against that -- allowing passive observers (including the mailbox server) to learn many things about your protocol even though the *contents* remain encrypted an unview-able by the server.

security.md
```diff
@@ -0,0 +1,94 @@
...
+between the peers and thus be harder to detect. Note that suche a
+machine-in-the-middle attack would require either intercepting encrypted TLS
+traffic, or compromising the rendezvous server itself.
+
+Both cases can be mitigated by enabling the verifier feature, in which case
+even successful attacks can be detected.
+
+In general, this is the same threat model as with ordinary key exchanges, except
```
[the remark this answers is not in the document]
piegamesde: Replacing this with a Diffie-Hellman comparison, as it turns out there are more different kinds of key exchanges than I'd remembered.

security.md
piegamesde: Oh that's interesting, why? Because I see trust as an essential part of threat modeling and assessing individual security risks.

security.md
piegamesde: Yes you can always connect to a relay from behind Tor, but that's not what I'm talking about. The transit protocol specifies "Tor hints" which point to onion domains, of which the idea is to spin up a hidden service and send that over as a hint. In that case, there is a "direct" connection with respect to the Tor network. But from an IP perspective, the Tor network merely acts as a relay.

security.md
```diff
@@ -0,0 +1,94 @@
...
+making it a single point of failure (SPOF).
+
+While it is of course possible to incapacitate the rendezvous server using
+traditional means of DoS, so much work is not necessary: An attacker may
+simply connect to every nameplate with a random code, causing the key exchange
+to fail. There is a "list" command in the protocol which makes it easy to
+enumerate all nameplates to disrupt them, but even without it the name space is
+sufficiently small (by design, as we want short codes) to brute force.
+
+There are several ideas to defend against this, however the only one already
```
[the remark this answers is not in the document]
piegamesde: Are you talking only about rejected ideas, or also about ideas under consideration/planning that we might add in the future?

security.md
meejah: I think this was via Tor people I'm thinking of? You aren't actually "trusting" the server, so the "expose" I guess is trying to get at what's actually happening. That is, that you reveal your IP address to them -- I don't think "trust" enters into it per se. Maybe "reveal" is a good one too?

I believe the logic is that it doesn't matter if you "trust" the server, or the people running the server -- the important thing is what information you reveal. For example, even if they are trustworthy they can themselves be compromised or forced to reveal any information they have (subpoena, etc)

security.md
meejah: Oh, it actually spins up Onion services?? Cool .. I guess it's been a while since I looked in there very deeply.

I suppose you could consider that Tor is basically acting "as" the relay in that case -- although I guess you could always say Tor is "a relay" in that sense. You could also view those as "merely" direct connections (albeit via Tor) since the semantics of Onion domains give you assurance you're connecting to the right computer.

Okay, I guess it's fine as-is.

It might still be worth distinguishing between "use Tor, but still via the Transit" vs. "use only Tor"...? (The former thing is possible, right?)

security.md
```diff
@@ -3,23 +3,23 @@
...
-If the attacker *additionally* has the capability of intercepting and changing
-network traffic ("MitM"), they might be able to fake arbitrary interactions
-between the peers and thus be harder to detect. Note that suche a
-machine-in-the-middle attack would require either intercepting encrypted TLS
-traffic, or compromising the rendezvous server itself.
+If the attacker successfully guesses the code, they can attempt a
+machine-in-the-middle (MitM) attack. For this, they quickly reconnect with
+the same code in order to connect with the second peer. With that, they get a
+secure channel to each side while being able to read and modify the connection's
+content.
```
meejah, line 14: :+1:

security.md
piegamesde: > Oh, it actually spins up Onion services?

Actually I don't know ^^ I know for sure that the protocol supports this and that this definitely was the original plan. But to which extent it has been implemented in the Python client, no idea. (IMO the latter question does not matter that much because this document should mostly speak from a protocol perspective.)

> It might still be worth distinguishing between "use Tor, but still via the Transit" vs. "use only Tor"...? (The former thing is possible, right?)

Yes, I'll try to work that in

piegamesde:
I added a commit that deprecates the `current_cli_version` field in the `welcome` message from the server. Of course the server will continue to provide it anyways, but it has no value to any other client other than the Python implementation. Should it be desirable we can create a new replacement for the other clients in the future.

meejah:
I don't actually see anywhere in the current mailbox code where it sends a `current_cli_version` (all I did is searched for that string but don't find anything). Anyway, fine to mention it here as "deprecated" (as you point out it's not going to work as-is for multiple kinds of clients anyway).

index.md
```diff
@@ -0,0 +1,28 @@
+# The Magic Wormhole protocols
+
+This site contains all the documentation and specification of protocols related
+to Magic Wormhole, which are not specific to a single client or implementation.
+It assumes the reader is already familiar with the general Magic Wormhole concept.
+
+The most important component is the **rendezvous server**. There are two aspects
```
meejah, line 7: maybe add "aka mailbox server" to avoid confusion (while we're still .. using both?)

index.md
```diff
@@ -0,0 +1,28 @@
...
+to it: The [**server protocol**](./server-protocol.md) describes how two peers
+find each other and how they then can exchange low-bandwidth messages.
+Once the two peers are connected over the server, the
+[**client protocol**](./client-protocol.md) describes how they do establish a
+secure way of exchanging messages.
+
+Using the established low-bandwidth secure channel, both sides then negotiate a
+ecure high-bandwidth channel, called [**transit**](./transit.md). The transit
```
meejah, line 15: typo -> secure

index.md
```diff
@@ -0,0 +1,28 @@
...
+protocol describes how both sides establish a direct connection, how a **relay
+server** may be used as fallback, and the cryptography used to make that
```
meejah, line 17: maybe add "aka Transit server" (similar reasoning to above)

index.md
```diff
@@ -0,0 +1,28 @@
...
+connection secure.
+
+Applications make use of the above protocols to provide their functionality.
+Currently, only one application level protocol is documented here:
+[**file transfer**](./file-transfer-protocol.md). Additionally, a custom
+[**uri scheme**](./uri-scheme.md) has been standardised for file transfer. This
+makes it possible for applications to replace the traditional code exchange
+with sharing a link or QR code.
+
+Security thread models and privacy considerations are discussed in
```
meejah, line 27: typo: thread -> threat

security.md
```diff
@@ -0,0 +1,97 @@
...
+Diffie-Hellman (or variations), except that MitM attackers additionally have to
+guess the code correctly. (Hence the name, "*password authenticated* key
+exchange", PAKE.)
+
+### Password entropy
+
+Compared to password strengths we are used to nowadays, 16 bits may seem scarily
+small, even knowing that an attacker only has one try to guess the password. But
+to put it into perspective: As a user, you'd have to experience 32000 failed
+transfers because of an attack in order to having a 50% chance of a file being
```
meejah, line 29: typo/grammar: "having" -> "have"

security.md
```diff
@@ -0,0 +1,97 @@
...
+intercepted. From an attacker's perspective, guessing thousands of codes is not
+difficult, but there is no possibility of targetting any individual connection,
+making any gained data fairly useless. From a service perspective, any successful
+attack implies 65536 failed attempts on average, which will quickly be noticed
```
meejah, line 33: I think you mean ~32000 here (to match above)...?

security.md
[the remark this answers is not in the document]
meejah: After more thought it's fine to just leave them out; we can add more if / when there's more-concrete plans etc.

security.md
```diff
@@ -0,0 +1,97 @@
...
+
+There are several ideas to defend against this, however the only one already
+specified in the protocol is the "permission" feature: The rendezvous server
+may use it to restrict new connections with some kind of challenge (or for
+private servers, authentication). The motivation behind this is to enable
+a proof-of-work challenge (using HashCash) when under attack to slow it down.
+See the server protocol specification for details.
+
+Currently the best defense against this is to use high-entropy nameplates which
+cannot easily be enumerated at the cost of longer codes. This has not be
```
meejah, line 63: If we're going to leave out speculation, maybe this too? (I'd have to do a lot more thinking to convince myself this is "the best" for example). Also typo: "This has not be" -> "This has not been"

security.md
```diff
@@ -0,0 +1,97 @@
...
+including the IP address of any two peers.
+
+Additionally, in order to establish a direct connection, public and local IP
+addresses are sent to the peer. This may leak information about the local
+network topology, including potentially used VPNs. If you do not trust your
+peer with such information, enforcing a relay server connection is advised.
+However, keep in mind that it is easy to host a relay server, so it might still
+leak your public IP address.
+
+If you do not want to expose your IP address to any of the servers, use
```
meejah, line 91: Maybe "any of the computers"? (To me in context "server" here means mailbox or transit .. but we want to make clear that you can avoid exposing your IP to the peer too).

server-protocol.md
```diff
@@ -72,19 +72,18 @@ messages. Clients must ignore unrecognized message types from the Server.
 
 The first thing the server sends to each client is the `welcome` message.
 This is intended to deliver important status information to the client that
-might influence its operation. The Python client currently reacts to the
-following keys (and ignores all others):
+might influence its operation. Clients should look out for the following fields,
+and handle them accordingly, if present:
```
meejah, line 76: Maybe add "clients should simply ignore other fields they don't understand"...?

meejah (approved):
Overall looks good .. some inline typo/other fixups.

All the new words include "rendezvous" and "relay" which I thought we agreed to avoid in favour of "mailbox" and "transit" (or "transit relay") respectively..? At the very least we should mention the other terms to avoid confusion I feel like.

security.md
meejah: (At least on the Python side I'm pretty sure we _don't_ enumerate local IPs etc when Tor mode is turned on...? I'd have to double-check that, but that's what it should do)

security.md
piegamesde: Yes, we can leave it out for now. But actually I have some old proposal for "high entropy nameplates" lying around, I could dig it out again and do this, should not be too hard in general.

server-protocol.md
piegamesde: I thought this was kind of implicit somehow, but no harm done in stating it explicitly.

security.md
meejah: If users can exchange high-entropy nameplates, can't they just exchange a key directly?
(I guess I'm assuming "high entropy" here means "32 bytes" ...)

server-protocol.md
meejah: :+1: yeah, "explicit usually better" IME.

felinira (approved):
Thanks for cleaning this up

security.md
piegamesde: Nope, should be correct. The one is the total average, the other one is only up to 50% probability. Something something Birthday paradox
(The probability of having a cracked password after 65536 attempts is less than one, but the probability of needing less than that also exists so it exactly cancels out on average)

meejah: Can you explain further? I don't immediately see how the Birthday Paradox is relevant here?

piegamesde: Okay thought about this more and it's not the Birthday paradox. So IMO 1/65536 still is correct, but one needs a different calculation for the other number.

I end up at (1 − 1/65536)^45000 = 0.503258791 so if the calculation is correct one would need 45000 guesses for a 50% success probability.

Fxrh: After 32768 attempts, you have an *expected* number of successful guesses that is ``0.5``. However, this number also includes events where you guess the right key multiple times, which doesn't really help you as an attacker. Therefore, the number of attempts you need to have a 50% chance of getting *at least* one correct guess is higher (and seems to be around ``45000``, as @piegamesde showed).

piegamesde: Addendum: this still works even in our model where the attacker only gets a guess at each code. In that case, I think it becomes "the probability of (not) being hacked after 45000 failed exchanges is 0.5"

So in the end I'd say both numbers are correct, they just answer different questions.

meejah: @Fxrh that part I don't understand: there's only one guess available per setup attempt, so what does "guess the right key multiple times" mean here..?

piegamesde:
Unless anybody has strong objections, I'm inclined to merge this now. As this is about documentation and does not make any substantial changes to the protocol, I don't think this requires doing an FCP and stuff.

security.md
meejah: I think the text could still benefit from more words, then? (And 65536 isn't either of these numbers, so that should change, yes?)

meejah:
I haven't re-reviewed since the force-push, and I think at least the 65536 number should be changed right? (to 32k or 45k depending on how you word it?)

piegamesde:
The force-push only added " at least once" as a clarification, see https://github.com/magic-wormhole/magic-wormhole-protocols/compare/95a1e65b510c88ebf552cd057ed27de26011456e..1286cb4159a048918bd2cfdff54708162da7d0ca.

The 65536 should be correct, because it is an average/expected value. (Okay actually it should be 65535, but that's not your point.) After 32768 attempts you'd expect 0.5 successes *on average*, which is not the same as a 50% probability for a single success.

meejah:
> The force-push only added " at least once" as a clarification, see https://github.com/magic-wormhole/magic-wormhole-protocols/compare/95a1e65b510c88ebf552cd057ed27de26011456e..1286cb4159a048918bd2cfdff54708162da7d0ca.

It would be a _lot_ easier to see this without force-pushes ...

> The 65536 should be correct, because it is an average/expected value. (Okay actually it should be 65535, but that's not your point.) After 32768 attempts you'd expect 0.5 successes _on average_, which is not the same as a 50% probability for a single success.

I'm more confused then, because the text says "any successful attack implies 65536 failed attempts on average, which will quickly be noticed" but the discussion justifies 32k or 45k. Surely on average there are less failures than that? (Or, please explain in the text).

piegamesde:
How about "From a perspective of the service, it will see on average one success and 65535 failures for every 2^16 attempts. Therefore, brute force attacks are inherently very disruptive of the service, and thus easy to detect."

> Surely on average there are less failures than that?

No, the chance of guessing a password with 16 bits strength in a single attempt is precisely 2^-16.

wuan:
We could simply say that an attacker would need to do more than 300k attempts to be at least 99% sure to have one successful attempt.

2^16 attempts would only lead to 63% probability, which is not really high.

maralorn:
> We could simply say that an attacker would need to do more than 300k attempts to be at least 99% sure to have one successful attempt (Or more than 196k for at least 95%).

I wouldn’t recommend that. This looks like we are putting the bar too high. Like, “Yeah, they might hack the connection. But they can’t be sure to definitely hack the connection, so we are fine …”

wuan:
> I wouldn’t recommend that. This looks like we are putting the bar too high. Like, “Yeah, they might hack the connection. But they can’t be sure to definitely hack the connection, so we are fine …”

An attacker who wants to have an acceptable probabilty (> 95%) to succeed will need to do that many attempts.

meejah:
> There is an extremely small probability to guess the password already on the initial attempt. This could be overcome by adding a verification message from the receiver back to the sender which can be used to identify the receiver before the data is transmitted.

This feature already exists.

meejah:
The chance of rolling a "1" on a d6 is 1/6 .. but we don't say that "in order to roll a 3 there will be on average 5 failures". (Or at least, this framing isn't familiar to me  ... but statistics classes were a *long* time ago ;).

Put another way: the attacker will not keep guessing once they have succeeded, so *on average* they will only need to try half that number of times. Only the least-lucky attacker will have 65535 failures first (well, a strictly _least_ lucky attacker will have an infinite number of failures if we channel H2G2 I suppose because they're so unlucky they will never succeed ;) ).

Anyway, your last suggestion sounds more clear (at least it highlights that the service should see a lot of failures if someone is doing a lot of guessing). I still feel like it's giving an incorrect impression that an attacker will *definitely* steal one connection after 65k tries (or that they need to on average try 65k times to get one, which they don't). That is, by your own reasoning from the "failure" side it should be 45425 (i.e. the first time the chance of failure goes below 50%) so why are you now using 65k here? If you have different reasoning, please respond in the thread about that.

(But, if you wanted to look from the failure way, after 65536 guesses there's still a decent chance of not getting any correct: if we insist on the attacker being unsuccessful every time, they have a (65535/65536) ^ 65536 = 0.37 chance of continuing to be unsuccessful after 65k guesses).

No matter what, I think the text should include some math at least as a footnote to justify whatever number is put here, and why. The important takeaway is that an attacker would have to do "a lot" of guesses.

piegamesde:
> but we don't say that "in order to roll a 3 there will be on average 5 failures"

the correct phrase here would be "for every 3 roll there will be on average 5 failures".

Anyways, I've reworked the paragraph again. I removed the 45000 number, because it doesn't help that it's correct if it's not intuitive to interpret what it represents, and the new version still gets the relevant point across IMO.

PR #33, opened 2023-03-14
Try to reword Dilation specification

piegamesde:
At the moment it is written more as the documentation of the Python implementation. In the end, we want it to be a standalone document that helps others implement that protocol too, without having to know any of the Python internals or Twisted.


dilation-protocol.md
```diff
@@ -9,34 +9,44 @@ must select the best one to use, and cleanly shut down the others. To make
...
+## Overview
+
+A dilated Wormhole connection involves several moving parts.
+
+The Mailbox on the rendezvous server remains accessible to the application.
+It is furthermore used by the Dilation implementation to exchange internal
+coordination messages.
+
+The Transit connection (TODO rename) is now managed by the Dilation
+implementation. It will automatically be established in the beginning.
```
vu3rdd, line 25: "It will automatically be established in the beginning..." looks very confusing to me. What exactly is meant here?

index.md
```diff
@@ -0,0 +1,28 @@
...
+find each other and how they then can exchange low-bandwidth messages.
+Once the two peers are connected over the server, the
+[**client protocol**](./client-protocol.md) describes how they establish a
+secure way of exchanging messages.
+
+Using the established low-bandwidth secure channel, both sides then negotiate a
+secure high-bandwidth channel, called [**transit**](./transit.md). The transit
+protocol describes how both sides establish a direct connection, how a special
+relay server may be used as fallback, and the cryptography used to make that
+connection secure.
```
meejah, line 18: This would probably benefit from a mention of Dilation (given this is the dilation PR).

dilation-protocol.md
```diff
@@ -1,371 +1,280 @@
...
+Therefore, many applications use the Transit protocol to establish a high-bandwidth channel to use for the acutal data transfer.
+
+Dilation builds upon this concept, but tries to make it a first-class citizen that is also easy to use.
+The idea is that clients say that they want to "dilate" a Wormhole,
+and the Dilation protocol will take care of the rest to establish such a high-bandwidth connection automatically.
+
+Additionally, the protocol takes care of automatically re-establishing such a channel after a connection interruption,
+including re-sending all messages that got lost.
+Furthermore, it has the concept of sub-channels built in,
+and will take care of the sub-channel management and data multiplexing too.
```
meejah: This "kind of" gets at the "durability" aspect, but I feel that should be stated more clearly.

What you get from the subchannels is a stream of data that will _definitely_ be delivered to the other side even changing networks, etc, until one side gives up or both sides are disconnected from the mailbox for "too long".

dilation-protocol.md
```diff
@@ -1,371 +1,280 @@
...
+
+## Overview
+
+A dilated Wormhole connection involves several moving parts.
+
+The Mailbox on the Mailbox server remains accessible to the application.
+It is additionally used by the Dilation implementation to exchange internal coordination messages.
+
+The usual Transit connection is replaced by a Dilation connection, that is managed by the Dilation implementation.
+Compared to the Transit connection, it has additional features like resilience against connection interruptions and sub-channels.
```
meejah: I guess this hints at durability too, but similar to above doesn't _quite_ state that?

dilation-protocol.md
```diff
@@ -1,371 +1,280 @@
...
 ## Capability discovery
 
-The Wormhole protocol includes a `versions` message sent immediately after
-the shared PAKE key is established. This also serves as a key-confirmation
-message, allowing each side to confirm that the other side knows the right
-key. The body of the `versions` message is a JSON-formatted string with keys
-that are available for learning the abilities of the peer. Dilation is
-signaled by a key named `can-dilate`, whose value is a list of strings. Any
-version present in both side's lists is eligible for use.
+As for Transit, this is up to the application layer protocol to handle, preferably
```
meejah: No, it's up to the protocol. There are two different "versions" messages: one for "internal" use and one for application use.

dilation-protocol.md
```diff
@@ -1,371 +1,280 @@
...
+using the `version` phase.
+
+TODO: This section described the version exchange within the `version` phase,
+however there is also some version information in one of the first Dilation
+messages itself. These are more or less redundant, but which one should we take?
+ ~ piegames
```
meejah: I do not believe they are redundant.

dilation-protocol.md
```diff
@@ -1,371 +1,280 @@
...
+
+## Mailbox changes
+
+TODO maybe move to ./client-protocol.md? ~piegames
+
+The Mailbox on the rendezvous server is used to deliver dilation requests and connection hints.
+The current mailbox protocol uses named "phases" to distinguish messages
+(rather than behaving like a regular ordered channel of arbitrary frames or bytes),
+and all-number phase names are reserved for application data.
+Therefore the dilation control messages use phases named `DILATE-0`, `DILATE-`, etc.
```
meejah: typo (DILATE-1 I guess?)

dilation-protocol.md
```diff
@@ -1,371 +1,280 @@
...
+Dilation is initiazed by sending a `please` (i.e. "please dilate") type message with a set of versions that can be accepted.
+Versions use strings, rather than integers, to support experimental protocols, however there is still a total ordering of preferability.
 
 ```
-{ "type": "please",
+{
+  "type": "please",
   "side": "abcdef",
-  "accepted-versions": ["1"]
+  // "accepted-versions": ["1"] TODO that field is not part of the Python implementation ~piegames
```
meejah: Right, this is already communicated in the "versions" message isn't it?

dilation-protocol.md
```diff
@@ -1,371 +1,280 @@
...
-wouldn't have started gen-2).
-
-(TODO: what about a follower->leader connection that was started before
+If one side receives a `please` before having initiated dilation itself,
+the contents are stored in case it decides to do so in the future.
+Once both sides have sent and received a `please` message,
+the side determines whether it is the leader or the follower.
+~~Both sides also compare `accepted-versions` fields to choose the best mutually-compatible
+version to use: they should always pick the same one.~~
+TODO that field is not part of the Python implementation ~piegames
```
meejah: The Leader picks the version.
We don't need to add `accepted-versions` since there's already `use-version`.

dilation-protocol.md
```diff
@@ -1,371 +1,280 @@
...
+
+### Establishing a Dilation connection
+
+Then both sides begin the connection process by opening listening sockets and sending `connection-hint` messages for each one.
+After a slight delay they will also open connections to the Transit Relay of their choice and produce hints for it too.
```
meejah: Can be multiple relays.

dilation-protocol.md
```diff
@@ -1,371 +1,280 @@
...
+The receipt of inbound hints (on both sides) will trigger outbound connection attempts.
+The hints are encoded as described in the Transit protocol.
+(TODO specify the exact encoding, give examples ~piegames)
+
+Some of these connections may succeed, and the Leader decides which to use (via an in-band signal on the established connection).
```
meejah: Mention what signal, probably.

dilation-protocol.md
```diff
@@ -1,371 +1,280 @@
...
+The others are dropped.
```
meejah: Tell peers what to do here.

dilation-protocol.md
```diff
@@ -1,371 +1,280 @@
...
+or after the Follower thinks the connection has been lost.
+In all cases, the Leader is the only side which should send `reconnect`.
+
+```json
+{ "type": "reconnect" }
+```
+
+Upon receiving a `reconnect`, the Follower should stop any pending connection attempts and terminate any existing connections
+(even if they appear viable).
+Listening sockets may be retained, but any previous connection made through them must be dropped.
```
meejah, line 149: There's some waffling here in the other doc, no? Because this doesn't necessarily produce a clear separation between generations?

dilation-protocol.md
```diff
@@ -1,371 +1,280 @@
...
+
+Once all connections have stopped, the Follower should send a `reconnecting` message,
+then start the connection process for the next generation,
+which will send new `connection-hint` messages for all listening sockets.
+
+```json
+{ "type": "reconnecting" }
+```
+
+The Leader will drop all existing connections from before sending the `reconnect`,
```
meejah: will -> must

dilation-protocol.md
```diff
@@ -1,371 +1,280 @@
...
-`reconnecting` and switches back to "connecting".
-
-"Connection hints" are type/address/port records that tell the other side of
-likely targets for L2 connections. Both sides will try to determine their
-external IP addresses, listen on a TCP port, and advertise `(tcp,
-external-IP, port)` as a connection hint. The Transit Relay is also used as a
-(lower-priority) hint. These are sent in `connection-hint` records, which can
-be sent any time after both sending and receiving a `please` record. Each
-side will initiate connections upon receipt of the hints.
+derived key ~warner — please please don't ~piegames)
```
meejah, line 169: Why not?
A major advantage of including this in the key derivation is that it makes it _impossible_ to confuse other-generation messages with _this_ generation.

dilation-protocol.md
```diff
@@ -1,371 +1,280 @@
...
 
-```
-{ "type": "connection-hints",
-  "hints": [ ... ]
-}
-```
+(TODO: reduce the number of round-trip stalls here, I've added too many ~warner)
+(yes, and one could also drastically reduce the number of messages by sending
+all hints at once with the reconnect message ~piegames)
```
meejah: There would need to be at least two separate ones: "not-relayed" and "relayed". (There's an intentional delay for relayed connections so that non-relayed are definitely preferred).

I think adding this requirement will slow the protocol down, because then you can't send any message until _all_ possible transports are set up. Tor transports (for example) can takes "dozens of seconds" to establish.

Overall it seems far more straightforward and flexible to have "N" hints than to try and reduce them. Or at least, I don't see what the point is in having "exactly 2" such messages with many hints in each versus a message per hint. You can already put multiple hints in one message (right?) so if the peer has many ready to go they can be batched within the current protocol.

dilation-protocol.md
```diff
@@ -1,371 +1,280 @@
...
-master PAKE key using HKDF. Each L2 connection uses the same dilation key,
-but different ephemeral keys, so each gets a different session key.
+TODO delegate this to Transit
+
+## Dilation connection
+
+TODO document the special relay handshake (it's the same as in Transit) ~piegames
+
+TODO How does this part of the protocol interact with web clients and WebSocket
+connections? ~piegames
```
meejah: I don't understand the question.

dilation-protocol.md
```diff
@@ -1,371 +1,280 @@
...
+Hints can arrive at any time. One side might immediately send hints that can be computed quickly,
+then send additional hints later as they become available.
+For example, it might enumerate the local network interfaces and send hints for all of the LAN addresses first,
+then send port-forwarding (UPnP) requests to the local router.
+When the forwarding is established (providing an externally-visible IP address and port),
+it can send additional hints for that new endpoint.
+If the other peer happens to be on the same LAN,
+the local connection can be established without waiting for the router's response.
 
+(I'd like to see that feature removed, this stuff is already complicated enough with a signle hints message ~piegames)
```
meejah: There's never a _single_ hints message...?

dilation-protocol.md
```diff
@@ -1,371 +1,280 @@
...
+which might have a unidirectional failure mode, at which point we'll need to
+monitor full roundtrips. To accomplish this, the Leader will send periodic
+unconditional PINGs, and the Follower will respond with PONGs. If the
+Leader->Follower connection is down, the PINGs won't arrive and no PONGs will
+be produced. If the Follower->Leader direction has failed, the PONGs won't
+arrive. The delivery of both will be delayed by actual data, so the timeouts
+should be adjusted if we see regular data arriving.~~
+
+If the Leader loses the Dilation connection for whatever reason, it will initiate a reconnection attempt.
+This involves sending a `reconnect` message and according hints over the Mailbox connection,
```
meejah, line 277: should mention generations / DILATE-n messages probably.

dilation-protocol.md
```diff
@@ -1,371 +1,280 @@
...
+The rest of the message depends upon the type:
+
+* 0x00 PING, 4-byte ping-id
+* 0x01 PONG, 4-byte ping-id
+* 0x02 OPEN, 4-byte subchannel-id, 4-byte seqnum
+* 0x03 DATA, 4-byte subchannel-id, 4-byte seqnum, variable-length payload
+* 0x04 CLOSE, 4-byte subchannel-id, 4-byte seqnum
+* 0x05 ACK, 4-byte response-seqnum
+
+TODO it would make more sense to have the seqnum before the subchannel-id, processing wise ~piegames
```
meejah: Why? Also you have to deserialize all of this to process it, so I can't see why order matters at all.

dilation-protocol.md
```diff
@@ -388,158 +297,53 @@ In addition, the Leader must:
...
-control-channel client endpoints can only be used once, and does not close
-until the Wormhole itself is closed.
-
-Each OPEN/DATA/CLOSE message is delivered to the L4 object for queueing,
-delivery, and eventual retirement. The L5 layer does not keep track of old
-messages.
+The DATA payloads are multiplexed over some virtual sub-channels. They may be
+exposed to the application as logically independent data streams.
+
+Each subchannel uses a distinct subchannel-id, which is a big-endian four-byte identifier.
```
meejah, line 346: Should include some mention that subchannel-ids are a "protocol" thing and *should not* be leaked through the application API.

(That is, application shouldn't be allowed to select or depend upon subchannel IDs).

dilation-protocol.md
```diff
@@ -388,158 +297,53 @@ In addition, the Leader must:
...
+Subchannel-ids should not be reused
+(it would probably work, the protocol hasn't been analyzed enough to be sure).
+
+The side that opens a sub-channel is named the Initiator, and the other side is named the Acceptor.
+Subchannels can be initiated in either direction, independent of the Leader/Follower distinction.
+For a typical file-transfer application, the subchannel would be initiated by the side seeking to send a file.
 
 ### Flow Control
 
+TODO review and potentially remove? ~piegames
```
meejah, line 369: I would say: absolutely not, flow-control is pretty central to streaming applications.

dilation-protocol.md
```diff
@@ -1,371 +1,280 @@
...
-The L3 manager monitors the viability of the current connection, and declares
-it as lost when bidirectional traffic cannot be maintained. It uses PING and
-PONG messages to detect this. These also serve to keep NAT entries alive,
-since many firewalls will stop forwarding packets if they don't observe any
-traffic for e.g. 5 minutes.
+and must be processed in a specific order
+(the Leader must not accept the Follower's message until it has generated its own).
+Noise allows handshake messages to include a payload, but we do not use this feature.
+
+All subsequent messages as known as "Noise transport messages", and are independent for each direction.
```
meejah: typo: as -> are

dilation-protocol.md
```diff
@@ -1,371 +1,280 @@
...
+Leader->Follower connection is down, the PINGs won't arrive and no PONGs will
+be produced. If the Follower->Leader direction has failed, the PONGs won't
+arrive. The delivery of both will be delayed by actual data, so the timeouts
+should be adjusted if we see regular data arriving.~~
+
+If the Leader loses the Dilation connection for whatever reason, it will initiate a reconnection attempt.
+This involves sending a `reconnect` message and according hints over the Mailbox connection,
+the same way as for initially establishing the connection.
+Once a new connection is established and the handshake completes,
+both sides re-transmit their outbound messages that were lost due to the connection interruption.
```
meejah: Careful here: these are L4 messages. If you instead re-transmitted L2 messages, the crypto may break (and another reason to use distinct keys for different generations).

meejah (changes requested):
I still see benefits in keeping layers distinct, or at least describing them that way. Naming aside, grouping the different concerns I think helps implementers -- knowing that the L4 layer doesn't have to care about anything to do with channel-selection, encryption, deserialization, framing etc. Basically all it is concerned with is the high-level messages and their acknowledgement.

We don't _have_ to strictly follow what's already there, but some separation of concerns is a big benefit I think. For example, you have a "todo" about ordering of subchannel vs seq-num serialization, but that doesn't matter because you've already gotten "one whole message" from L2 so it's irrelevant what order the bytes are in.

Have you read the latest draft of the python specification? Also it would be good to link that in this document.

(Some inline comments too of course :)

index.md
piegamesde: Actually, this change is not from this PR but already on master. It's just that I pushed branches in the wrong order, and GitHub hasn't gotten the memo yet. Therefore, I'd like to delay updating the `index.md` to a later PR.

(For reference, this PR officially starts after "[Merge branch 'main' into dilation](https://github.com/magic-wormhole/magic-wormhole-protocols/pull/33/commits/63507bfdda513f2ccf407711365412062193dfce)")

dilation-protocol.md
piegamesde: Yes. This is what I meant with "redundant", and I did not know which one to take. I'll remove it then.

dilation-protocol.md
piegamesde: Yes, has been resolved since writing it. (This was before WebSocket support in relays)

piegamesde:
Thanks for the review

> I still see benefits in keeping layers distinct, or at least describing them that way. […]
> We don't have to strictly follow what's already there, but some separation of concerns is a big benefit I think.

I don't like the concept of "layer" because to me it invokes feeling of an "OSI layers" stack, which is only true for parts of it. I'd rather think of them as interdependent components, and treat them as such. But you are right, the new wording could make the separation a bit more clear again.

One thing to note is that sometimes the boundaries are a bit fuzzy: For example while Python has L5 stacked on top of L4, one could also write the code that mostly does both at once, but delegates parts of it to a "subchannel manager". (Basically the reverse from what Python does.)

> Have you read the latest draft of the python specification? Also it would be good to link that in this document.

Not really no, these changes here are mostly a rebase from my work in last autumn. I think it has changes over there that I should incorporate here? (If possible, I'd like to delegate that too to a future PR)

dilation-protocol.md
piegamesde: Hm, what would you like to hear here?

dilation-protocol.md
piegamesde: To be clear, I meant "hints exchange". Of course there may be a lot of hints, but I was talking about each side sending only one message. Transit does it like that currently and it already is complicated enough.

dilation-protocol.md
piegamesde: I agree that it mostly doesn't matter, this is just me being pedantic.

dilation-protocol.md
piegamesde: Shouldn't it? I mean, I don't have any strong opinions on this, but the initial protocol said nothing about that so I'd like to think it through some more. Are there scenarios where applications might want to reference some communication channel for some reason? Does it hurt us to expose the channel number?

(Of course we can always add the restriction now to be sure and then lift it later on should there be need …)

dilation-protocol.md
piegamesde: Basically, as far as I can see flow control is something for the Dilation implementation to handle, and requires no protocol support at all. So if you remove the Python specifics from the paragraph, it will go something like "Implementations might offer some flow-control functionality or quality of service for the sub-channels, so that a single very busy channel does not starve the others. This is not a requirement though, as it significantly increases implementation complexity and most applications do just fine without".

dilation-protocol.md
piegamesde: > There would need to be at least two separate ones: "not-relayed" and "relayed".

Transit manages to do this while sending all hints at once, by delaying the connection attempt towards relay servers instead. This allows more flexibility in the implementation: For example, the Rust code connects to everything, but when a connection is a relay connection it waits a couple of seconds to see whether a "better" connection comes up.

> I think adding this requirement will slow the protocol down, because then you can't send any message until all possible transports are set up. Tor transports (for example) can takes "dozens of seconds" to establish.

This could be solved by booting up these connections up early (we already know whether the other side supports it as soon as we get their `versions` message).

meejah: The current spec _allows_ you to do this, doesn't it? (I mean: a "hints" message can already contain multiple hints). We shouldn't be too prescriptive here, there's just too many things going on and many, many possible network scenarios.

dilation-protocol.md
meejah: I don't see any advantage to this. (Or, how would the state-machine get any simpler?)

dilation-protocol.md
meejah: Yes, that's what Brian and I talked about some time back. I'm not sure if he wrote this down anywhere?

To me it does make sense, now; there are requirements on the numbers so they should be picked by the protocol. If an application does need to cross-reference channels somehow I think it makes way more sense for that application-protocol to add messages that make sense to that application. (For example, they could have an initial message indicating their name or ID or whatever).

meejah: (As you say we could always lift the prohibition later if that happens to make sense)

dilation-protocol.md
meejah: No, it would still say the stuff about pausing _ALL_ channels whenever a single one needs to provide backpressure. That is, that implementations _should not_ attempt to provide their own subchannel flow-control, and instead rely on TCP's. (Competing flow-controllers on the same channel do not play nicely)

meejah:
Yes, they're like the OSI stack: a "higher layer" can depend on a lower one, but _NOT_ vice-versa. This is a good thing.

So they're not "inter"-dependent. It's fine if you want to assign different names, I suppose, but for network protocol implementers the OSI stack is very familiar and talking about "layers" is pretty common in lots of Internet protocols.

It also lets you forget / compartmentalize complexity: I can forget about framing, encryption, ACKs, re-sends, etc etc at the higher layers and just think about the logical flow of in-order messages. It also helps with state-machine diagrams and separation.

dilation-protocol.md
piegamesde: Not the state machine per se, but other things like for example prioritizing hints and merging duplicate hints. Currently you can implement it as one task like "have some hints now go and get me a connection", but with this you need to account for a way to side-load new hints into the process.

dilation-protocol.md
piegamesde: This would require us to stringently define what *exactly* a generation is, and how to count them, and also to make sure that this definition does not cause desyncs on badly-timed connection interruptions.

Before we go that way, I'd like to take a step back and ask what are exactly the issues we are concerned about here?

dilation-protocol.md
piegamesde: Oh I see, I think I was confusing flow control and quality of service. (Tbf not entirely my fault, because the paragraph also talks about the one while actually describing the latter.)

So is the takeaway here "don't do your own flow control but feel free to implement QoS"?

dilation-protocol.md
meejah: I don't see any way to avoid that, and don't think it's desirable anyway as the state-machine will look approximately the same while making the application slower for users (and less flexible).

dilation-protocol.md
meejah: The original (https://github.com/magic-wormhole/magic-wormhole/blob/master/docs/dilation-protocol.md#flow-control) doesn't mention QoS at all, and I wouldn't mention it here.

piegamesde: I know but it *talks* about QoS: "e.g. a chat conversation sharing a wormhole with file-transfer might prefer the IM text to take priority". Therefore why not talk about that explicitly?

meejah: I guess that's certainly a use-case for QoS -- but what do you think is worth mentioning here? There's nothing to be done by implementers -- they have to pause "the TCP stream" to achieve backpressure.

If anything, this would be an interesting point for application developers: mixing high-bandwidth subchannels with low-bandwith ones may perform poorly under bad network conditions. But I think all they can do here is "not do that"?

That is, I don't believe that sentence is supposed to imply to Dilation implementers that they should attempt to implement some sort of QoS system. I don't recall details on how that'll interact (or not) with TCP windowing etc. (Usually QoS is done per-TCP-stream, e.g. prioritizing one whole stream over another).

dilation-protocol.md
meejah: Yes, generations are strictly defined and should be specified better if you don't think that's true.

dilation-protocol.md
meejah: (I see you resolved this, but I don't seen changes .. maybe your workflow is different from mine).

This should also say how a peer determines which messages were lost (or potentially lost).

piegamesde: Oops I'm sorry.

I usually keep track of open discussions as my TODO list (at least for things that are easily fixable), but I forgot to push my changes after doing so.

piegamesde: Unresolving because of "This should also say how a peer determines which messages were lost"

dilation-protocol.md
piegamesde: But what is the purpose of having a generation counter and using it somewhere in the key derivation? What problem does it solve that we would have otherwise?

PR #34, opened 2023-03-14
Dilation

piegamesde:
This is the pull request of the `dilation` branch, where we accumulate all our Dilation related changes.


dilation-protocol.md
```diff
@@ -24,28 +24,30 @@ version present in both side's lists is eligible for use.
 ## Leaders and Followers
 
 Each side of a Wormhole has a randomly-generated dilation `side` string (this
-is included in the `please-dilate` message, and is independent of the
```
meejah, line 27: Why are we deleting the "-dilate" from here..? Seems more clear to me to keep it.

piegamesde: See #26. It's to make the documentation match the existing code

dilation-protocol.md
```diff
@@ -24,28 +24,30 @@ version present in both side's lists is eligible for use.
...
-version of the intersection of the two sides' abilities as reported in the
-`versions` message. Both sides will use whichever `use-version` was specified
-by the Leader (they learn which side is the Leader at the same moment they
-learn the peer's `use-version` value). If the Follower cannot handle the
-`use-version` value, dilation fails (this shouldn't happen, as the Leader
-knew what the Follower was and was not capable of before sending that
-message).
+The version message payload has a key called `can-dilate` that has a
+list of versions supported by a particular version of the client (eg:
+["1", "2"]). After receiving the version message from the other
```
meejah, line 43: This kind of gets rid of how the negotiation actually happens .. can we leave the `use-version` and `versions` details (i.e. it no longer says how to compute the "shared" versions or who decides which version to use and how that's communicated.

dilation-protocol.md
meejah: Ah, okay sorry I thought it had "-dilate" in the code :+1:

dilation-protocol.md
piegamesde: I'm not sure I understand your point, sorry

PR #36, opened 2023-03-26
Server protocol improvements

piegamesde:
(This builds on top of the changes in #32 and thus depends on that being merged first.)

Numerous small changes to the server protocol that either better document existing behavior or change the semantics.

I tried to keep the diff as minimal while stating the intent of the changes, but the entire file is pending a rewrite and adding diagrams. I plan on doing that later, as I want to get agreement on the changes themselves first.

None of these changes should be client-breaking, except for the fact of sending `closed` before `close` for the mailbox half-closing mechanism. That, I am unsure about. Server side, some things will need to be double checked, but I hope it should be generally fine too.

### DIFF
--- a/server-protocol.md
+++ b/server-protocol.md
@@ -289,7 +289,7 @@ any), and which ones provoke direct responses:
 * (C->S) open {mailbox:}
 * (C->S) add {phase: str, body: hex} -> message (to all connected clients)
 * S->C message {side:, phase:, body:, id:}
-* (C->S) close {mailbox:?, mood:?} -> closed
+* (C->S) close {mailbox:?, mood:?} -> closed (the mailbox is known and implicit to the current connection)
 * S->C closed
 * S->C ack
 * (C->S) ping {ping: int} -> ping
--- a/server-protocol.md
+++ b/server-protocol.md
@@ -198,6 +198,10 @@ it (to avoid a race condition), but for simplicity, clients can send a claim for
 all nameplates, even ones which they've allocated themselves.
 Note that a claim (and therefore an allocation) also automatically opens the
 mailbox, see below.
+The `claim` command has the optional boolean field `allocate` (defaulting to
+`true`), which if set to `false` will result in an error if the to be claimed
+nameplate does not already exist. The intent behind this is clients who want
+to insure themselves against user mistakes when entering a nameplate.
 
 Nameplates (on the server) must live until the second client has learned
 about the associated mailbox, after which point they can be reused by other
@@ -291,7 +295,7 @@ any), and which ones provoke direct responses:
 * S->C nameplates {nameplates: [{id: str},..]} (response might be empty)
 * (C->S) allocate {} -> allocated
 * S->C allocated {nameplate:}
-* (C->S) claim {nameplate:} -> claimed
+* (C->S) claim {nameplate:, allocate:?bool} -> claimed
 * S->C claimed {mailbox:}
 * (C->S) release {nameplate:?} -> released
 * S->C released
--- a/server-protocol.md
+++ b/server-protocol.md
@@ -194,8 +194,10 @@ released it, or after some period of inactivity.
 
 Clients can either make up nameplates themselves, or (more commonly) ask the
 server to allocate one for them. Allocating a nameplate automatically claims
-it (to avoid a race condition), but for simplicity, clients send a claim for
+it (to avoid a race condition), but for simplicity, clients can send a claim for
 all nameplates, even ones which they've allocated themselves.
+Note that a claim (and therefore an allocation) also automatically opens the
+mailbox, see below.
 
 Nameplates (on the server) must live until the second client has learned
 about the associated mailbox, after which point they can be reused by other
@@ -232,6 +234,13 @@ are delivered immediately to the connected client. There is no explicit ack
 to the `open` command, but since all clients add a message to the mailbox as
 soon as they connect, there will always be a `message` response shortly after
 the `open` goes through. The `close` command provokes a `closed` response.
+Note that the `claim` and `allocate` commands implicitly already open the
+mailbox. The `open` command may be issued nevertheless for client implementation
+simplicity, and then simply be a no-op. There is also the possibility to open
+a mailbox directly without using a nameplate. This is mainly useful for
+reconnecting clients that had a mailbox previously, but may be used by future
+protocols too (see for example
+https://github.com/magic-wormhole/magic-wormhole-protocols/issues/2).
 
 The `close` command accepts an optional "mood" string: this allows clients to
 tell the server (in general terms) about their experiences with the wormhole
--- a/server-protocol.md
+++ b/server-protocol.md
@@ -230,7 +230,13 @@ coming from the other client.
 Each mailbox is "opened" by some number of clients at a time, until all
 clients have closed it. Mailboxes are kept alive by either an open client, or
 a Nameplate which points to the mailbox (so when a Nameplate is deleted from
-inactivity, the corresponding Mailbox will be too).
+inactivity, the corresponding Mailbox will be too). The mailbox server makes
+sure a mailbox can never be opened by more than two parties, neither
+simultaneously not cumulatively. Trying to open a mailbox as a third party
+results in a "crowded" error for that attempt. Moreover, when a client connects
+to a mailbox which the other client closed and thus has no chance of ever seeing
+a its peer, the server may reject its attempt to open the mailbox with an error
+message.
 
 The `open` command both marks the mailbox as being opened by the bound side,
 and also adds the WebSocket as subscribed to that mailbox, so new messages
--- a/server-protocol.md
+++ b/server-protocol.md
@@ -205,14 +205,12 @@ nameplates for that whole time.
 
 The `allocate` command allocates a nameplate (the server returns one that is
 as short as possible), and the `allocated` response provides the answer.
-Clients can also send a `list` command to get back a `nameplates` response
-with all allocated nameplates for the bound AppID: this helps the code-input
-tab-completion feature know which prefixes to offer. The `nameplates`
-response returns a list of dictionaries, one per claimed nameplate, with at
-least an `id` key in each one (with the nameplate string). Future versions
-may record additional attributes in the nameplate records, specifically a
-wordlist identifier and a code length (again to help with code-completion on
-the receiver).
+
+There is a `list` command (with the answer message being `nameplates`) intended
+for the use-case of listing currently in-use nameplates for user input
+auto-completion purposes. However, this feature could trivially be used to
+disrupt the service, therefore servers may send an always-empty response to not
+disclose any information about in-use nameplates.
 
 ## Mailboxes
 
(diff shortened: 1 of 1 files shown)


server-protocol.md
```diff
@@ -226,32 +234,23 @@ coming from the other client.
...
-  indicating that either the wormhole code was typed in wrong, or an attacker
-  tried (and failed) to guess the code
-* `errory`: the client encountered some other error: protocol problem or
-  internal error
-
-The server will also record `pruney` if it deleted the mailbox due to
-inactivity, or `crowded` if more than two sides tried to access the mailbox.
+Note that the `claim` and `allocate` commands implicitly already open the
+mailbox. The `open` command may be issued nevertheless for client implementation
+simplicity, and then simply be a no-op. There is also the possibility to open
```
meejah, line 250: It's required, because this is how the client learns the `mailbox_id`.

server-protocol.md
```diff
@@ -226,32 +234,23 @@ coming from the other client.
...
+a mailbox directly without using a nameplate. This is mainly useful for
+reconnecting clients that had a mailbox previously, but may be used by future
+protocols too.
```
meejah, line 253: I assume you mean "Seeds" here? We could link to a relevant ticket.

server-protocol.md
```diff
@@ -205,14 +215,12 @@ nameplates for that whole time.
...
-least an `id` key in each one (with the nameplate string). Future versions
-may record additional attributes in the nameplate records, specifically a
-wordlist identifier and a code length (again to help with code-completion on
-the receiver).
+
+There is a `list` command (with the answer message being `nameplates`) intended
+for the use-case of listing currently in-use nameplates for user input
+auto-completion purposes. However, this feature could trivially be used to
+disrupt the service, therefore servers now **must not** disclose any information
+about in-use nameplates. Clients should expect always getting an empty response.
```
meejah, line 223: I think it would be better (overall) to make clear what are unimplemented future features/directions versus current behavior. (That is, currently this is an operator decision, not a protocol decision .. although we may decide to deprecate this / add the wordlist thing -- so maybe link to a ticket here)

meejah: The only "must" here (currently) is that clients cannot depend on getting any information: they may (always) get an empty response to this request.

server-protocol.md
```diff
@@ -269,9 +268,37 @@ blob (for the VERSION message, as well as all application-provided payloads).
...
+* `errory`: the client encountered some other error: protocol problem or
+  internal error
+
+The server will also record `pruney` if it deleted the mailbox due to
+inactivity, or `crowded` if more than two sides tried to access the mailbox.
+
+A client sending a `close` message issues a `closed` message on the other client,
+to notify it about the intended session teardown. The latter client can then
+continue to send messages (so-called "half-closed state") until sending a `close`
+message too, which in turn will send the `closed` command to the first client and
```
meejah, line 295: Here, too, I think we should link to the "closing" ticket and come to a conclusion there..? Also we should make it clear here between "current behavior" and "proposed future change". How will a "new" client know if it is talking to an "old" client (that doesn't do this)?

(e.g. what's "a closed message"? Presumably you mean a special sort of number-phase message here? Or something else?)

server-protocol.md
```diff
@@ -280,18 +307,18 @@ any), and which ones provoke direct responses:
...
 * S->C allocated {nameplate:}
-* (C->S) claim {nameplate:} -> claimed
+* (C->S) claim {nameplate:, dont-allocate:?bool} -> claimed
 * S->C claimed {mailbox:}
-* (C->S) release {nameplate:?} -> released
-* S->C released
-* (C->S) open {mailbox:}
+* ~~(C->S) release {nameplate:?} -> released~~ (deprecated)
+* ~~S->C released~~
+* (C->S) open {~~mailbox:~~} (the mailbox is known and implicit to the current connection)
```
meejah, line 318: How does the client learn the mailbox-id then?

server-protocol.md
```diff
@@ -186,16 +186,26 @@ observers to use the hard-earned token for themselves.
...
+is expected to reconnect after a short amount of time.)
 
 Clients can either make up nameplates themselves, or (more commonly) ask the
 server to allocate one for them. Allocating a nameplate automatically claims
-it (to avoid a race condition), but for simplicity, clients send a claim for
+it (to avoid a race condition), but for simplicity, clients can send a claim for
 all nameplates, even ones which they've allocated themselves.
+Note that a claim (and therefore an allocation) also automatically opens the
+mailbox, see below.
+The `claim` command has the optional boolean field `dont-allocate`, which if set
```
meejah, line 205: Might be easier to read if this is flipped around (i.e. the boolean is `allocate` with implicit default of `true`).

server-protocol.md
```diff
@@ -186,16 +186,26 @@ observers to use the hard-earned token for themselves.
...
-On the Rendezvous Server, the Nameplate contains a pointer to a Mailbox.
-Clients can "claim" a nameplate, and then later "release" it. Each claim is
+On the Mailbox server, the Nameplate contains a pointer to a Mailbox.
+Clients can "claim" a nameplate, each claim is
 for a specific side (so one client claiming the same nameplate multiple times
-only counts as one claim). Nameplates are deleted once the last client has
-released it, or after some period of inactivity.
+only counts as one claim). There used to be an explicit `release` command, but
+Nameplates are new implicitly deleted once both clients have claimed
+it and thus found their mailbox, or after some period of inactivity. A client
```
meejah, line 194: Do clients still learn explicitly when they definitely have (or don't) a nameplate?

meejah (changes requested):
Note that you _can_ make "a PR against another PR" basically (if you want to just see the relevant changes here).

This lumps together several changes and might be better served with individual tickets? e.g. come to a conclusion on the "closing" stuff, and _then_ update the spec. Similar for "released", "list" etc.

I haven't thought hard about the backwards / forwards compatibility concerns of all these changes, but don't see any discussion about it: that is, what happens when a new client is on one side and an old one on the other? What about a new client talking to an "old" server? Do they need to know and behave differently? That is, there are several cases to consider: old vs new server and matched new, matched old, or mismatched peers.

It's in one comment, but I also think the spec could benefit from making more clear what are current behaviors vs. "planned future changes". Ideally by linking to tickets (especially because without tickets on the server, these changes may never happen ;) ).

server-protocol.md
meejah: (Hmm, re-reading this and it's not clear to me who "issues a closed message": do you mean the server does this? Or the other client?)

Anyway, I think this is best concluded in the existing ticket about graceful session shutdown: https://github.com/magic-wormhole/magic-wormhole-mailbox-server/issues/31

meejah:
p.s. overall it's great to move this stuff forward, but I don't want to end up with a "spec" that has zero implementations -- this is partly why I guess I'm suggesting separate tickets, or at least words that indicate the status (and link to e.g. "implement this in the server" tickets).

server-protocol.md
piegamesde: Yeah I've thought of that too. Either a client doesn't, which technically works fine until it has to reconnect … or we add the `mailbox` to the `claimed` response, or we sent an `opened` message as second answer to the `claim` command as if there had been an `open`. I'm unsure about backwards compatibility in the third case, but the second one would definitely work.

server-protocol.md
piegamesde: Thought of this too, the idea was that `false` is easier as default value. I like it either way around

server-protocol.md
piegamesde: How do you mean?

piegamesde:
Thanks for the feedback.

I'll go ahead and start implementing stuff server side soon. Also I'll remove the "session shutdown" related changes as they clearly require some more discussion.

server-protocol.md
piegamesde: I see two options here: either we say servers **should not** disclose in-use nameplates, in which case the respective server configuration its default value flipped. Or we make it a **must not**, in which case it may be deprecated/removed. Either way, nothing really changes for clients unless they (wrongly) assume to always get a response with content.

One more question would be whether the server should give an error instead of an empty response, although in the end it does not matter that much.

meejah: I'm not opposed to changing the behavior in the future, but _right now_ this is up to the operator. So these words should reflect that (possibly linking to a server ticket describing the changed / future behavior).

Both currently-deployed public servers allow `LIST`.

server-protocol.md
piegamesde: When a client connects and has no chance of ever finding a peer (as is the case in the described scenarios), the best thing is to give it an error message when it tries to `claim` the nameplate. The alternative is to just free the nameplate and let the client go stale, which is the status quo AFAICT.

meejah: Yes, I think proposing something to fix that hole could be good .. what I'm getting at here, though, is that the above words change the protocol from "explicit" (you must `RELEASE` the nameplate) to "implicit" so we need to make sure the clients still know the status of their nameplate.

piegamesde: You mean that old clients may expect the nameplate to exist until they explicitly send "release"? That's a valid concern, however I don't see how this information could make any change in behavior that may be relevant here. Because once they've connected to the mailbox, they don't care about the nameplate anymore anyways.

Do you have some scenario in mind where my idea might go wrong?

meejah: If a client thinks it still "has" a nameplate, and gets disconnected in that state, it will reconnect and try to "claim". I haven't analyzed the above to learn if that _could_ happen here though.

Regardless, it's often better "in general" to be explicit and consistent, is why I'm pursing this angle.

piegamesde: The problem with "explicit" here is that we must wait until clients release it, which they currently only do until after the encryption handshake for some reason, which leaves us open to situations like in https://github.com/magic-wormhole/magic-wormhole-mailbox-server/issues/19

As far as I understand, a client, once ever claimed a nameplate, should always reconnect with `open` only. Should this not be the case in the Python implementation, then I'll change the wording and server side state machine to track this so that it cannot go wrong in the way you described, while still preventing the race hazard I'm trying to fix.

piegamesde: > then I'll change the wording and server side state machine to track this so that it cannot go wrong in the way you described, while still preventing the race hazard I'm trying to fix.

After three hours of trying, it turns out that this is a lot more difficult than it should be. The problem is that after opening a mailbox, clients don't know who else is there or not until messages have been sent.

So either we come to the conclusion that your concern here is a non-issue and keep the current proposal, or I'll remove that bit and try to tackle it again once we do proper session closing.

piegamesde:
I've removed quite a few of the proposed features, as they turned out to be more complicated than anticipated. There is now a pull request at the server repository, which should implement all the changes proposed in here: https://github.com/magic-wormhole/magic-wormhole-mailbox-server/pull/38

meejah:
Upon re-reading, I do not understand why we'd want to change the protocol to make CLAIM _not_ allocate a mailbox-id? Why would a client want to CLAIM a nameplate, but not allocate a mailbox-id? If they _did_ do a CLAIM with "allocate=False", how do they allocate a mailbox-id?

If there are scenarios that these changes are intended to fix, can you either file or link to relevant tickets please?

piegamesde:
> The intent behind this is clients who want to insure themselves against user mistakes when entering a nameplate.

this is for the second party, under the assumption that a mailbox for that nameplate has already been allocated.

meejah:
> > The intent behind this is clients who want to insure themselves against user mistakes when entering a nameplate.
> 
> this is for the second party, under the assumption that a mailbox for that nameplate has already been allocated.

I see, so you mean for a second party that has typed the code in wrong (with the incorrect nameplate)?
Sorry I must have re-read that part too fast :) 

(Regardless of the outcome of this "protocol" PR, are there tickets filed anywhere else that we can refer to?)

